PR #3, opened 2021-01-27
WIP experimental pointwise dot() for powerlaw Legendre

TSGut:
The aim is to allow for fast and importantly t->1 compatible computation of \int_{-1}^1 (t-x)^a f(x) g(x) dx where f and g are given in Legendre polynomial coefficients.

### DIFF
--- a/src/stieltjes.jl
+++ b/src/stieltjes.jl
@@ -86,72 +86,31 @@ const PowKernelPoint{T,V,D,F} =  BroadcastQuasiVector{T, typeof(^), Tuple{Contin
 # cached operator implementation
 ####
 # Constructors support BigFloat and it's recommended to use them for high orders.
-mutable struct PowerLawIntegral{T, PP<:AbstractQuasiMatrix} <: AbstractCachedMatrix{T}
-    P::PP # OPs - only Legendre supported for now
-    a::T  # naming scheme follows (t-x)^a
-    t::T
-    data::Matrix{T}
-    datasize::Tuple{Int,Int}
-    array
-    function PowerLawIntegral{T, PP}(P::PP, a::T, t::T) where {T, PP<:AbstractQuasiMatrix}
-        new{T, PP}(P,a,t, pointwisecoeffmatrixdense(a,t,10),(10,10))
-    end
-end
-PowerLawIntegral(P::AbstractQuasiMatrix, a::T, t::T) where T = PowerLawIntegral{T,typeof(P)}(P,a,t)
-size(K::PowerLawIntegral) = (∞,∞) # potential to add maximum size of operator
-mutable struct PowerLawMatrix{T, PP<:AbstractQuasiMatrix} <: AbstractCachedMatrix{T}
-    P::PP # OPs - only Legendre supported for now
+mutable struct PowerLawMatrix{T, PP<:Normalized{<:Any,<:Legendre{<:Any}}} <: AbstractCachedMatrix{T}
+    P::PP # OPs - only normalized Legendre supported for now
     a::T  # naming scheme follows (t-x)^a
     t::T
     data::Matrix{T}
     datasize::Tuple{Int,Int}
     array
     function PowerLawMatrix{T, PP}(P::PP, a::T, t::T) where {T, PP<:AbstractQuasiMatrix}
-        new{T, PP}(P,a,t, pointwisecoeffmatrixdense(a,t,10),(10,10))
+        new{T, PP}(P,a,t, gennormalizedpower(a,t,10),(10,10))
     end
 end
 PowerLawMatrix(P::AbstractQuasiMatrix, a::T, t::T) where T = PowerLawMatrix{T,typeof(P)}(P,a,t)
 size(K::PowerLawMatrix) = (∞,∞) # potential to add maximum size of operator
 
 # data filling
-cache_filldata!(K::PowerLawIntegral, inds) = fillcoeffmatrix!(K, inds)
 cache_filldata!(K::PowerLawMatrix, inds) = fillcoeffmatrix!(K, inds)
 
 # because it really only makes sense to compute this symmetric operator in square blocks, we have to slightly rework some of LazyArrays caching and resizing
-function getindex(K::PowerLawIntegral{T, PP}, I::CartesianIndex) where {T,PP<:AbstractQuasiMatrix}
+function getindex(K::PowerLawMatrix{T, PP}, I::CartesianIndex) where {T,PP<:AbstractQuasiMatrix}
     resizedata!(K, Tuple(I))
     K.data[I]
 end
-function getindex(K::PowerLawIntegral{T,PP}, I::Vararg{Int,2}) where {T,PP<:AbstractQuasiMatrix}
-    resizedata!(K, Tuple([I...]))
-    K.data[I...]
-end
-function getindex(K::PowerLawMatrix{T, PP}, I::CartesianIndex) where {T,PP<:AbstractQuasiMatrix}
-    resizedata!(K, Tuple(I))
-    ℓ = K.datasize[1]
-    return (InfiniteArrays.Diagonal((2 .* (0:ℓ-1) .+1)/2)*(K.data))[I]
-end
 function getindex(K::PowerLawMatrix{T,PP}, I::Vararg{Int,2}) where {T,PP<:AbstractQuasiMatrix}
     resizedata!(K, Tuple([I...]))
-    ℓ = K.datasize[1]
-    return (InfiniteArrays.Diagonal((2 .* (0:ℓ-1) .+1)/2)*(K.data))[I...]
-end
-function resizedata!(K::PowerLawIntegral, nm) 
-    olddata = K.data
-    νμ = size(olddata)
-    nm = (maximum(nm),maximum(nm))
-    nm = max.(νμ,nm)
-    nm = (maximum(nm),maximum(nm))
-    if νμ ≠ nm
-        K.data = similar(K.data, nm...)
-        K.data[axes(olddata)...] = olddata
-    end
-    if maximum(nm) > maximum(νμ)
-        inds = Array(maximum(νμ):maximum(nm))
-        cache_filldata!(K, inds)
-        K.datasize = nm
-    end
-    K
+    K.data[I...]
 end
 function resizedata!(K::PowerLawMatrix, nm) 
     olddata = K.data
@@ -174,12 +133,12 @@ end
 ####
 # methods
 ####
-function *(K::PowKernelPoint,Q::Legendre{T}) where T
+function *(K::PowKernelPoint,Q::Normalized{<:Any,<:Legendre{<:Any}})
     t = K.args[1][0.]
     a = K.args[2]
     return Q*PowerLawMatrix(Q,a,t)
 end
-function dot(f::AbstractVector{T}, K::PowerLawIntegral, g::AbstractVector{T}) where T
+function dot(f::AbstractVector{T}, K::PowerLawMatrix, g::AbstractVector{T}) where T
     (i,conv1,conv2) = (0,1,2)
     while abs(conv1-conv2)>1e-15
         i = i+1
@@ -188,7 +147,7 @@ function dot(f::AbstractVector{T}, K::PowerLawIntegral, g::AbstractVector{T}) wh
     end
     return conv2
 end
-function *(g::Adjoint, K::PowerLawIntegral, f::AbstractVector)
+function *(g::Adjoint, K::PowerLawMatrix, f::AbstractVector)
     return dot(g',K,f)
 end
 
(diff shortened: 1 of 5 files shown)


TSGut:
@dlfivefifty I'm starting this PR to keep track of progress on this. 

For now, only a very straightforward pointwise evaluation via recurrence is implemented and there are almost certainly still big clashes with the quasi arrays design philosophy.

src/stieltjes.jl
```diff
@@ -75,4 +75,119 @@ end
...
+        i = i+1
+        conv1 = pointwisedot(f[1:i*20],g[1:i*20],a,t)
+        conv2 = pointwisedot(f[1:2*i*20],g[1:2*i*20],a,t)
+    end
+    return conv2
+end
+function dot(f::AbstractVector,M::PowerLawMatrix,g::AbstractVector)
+    return dot(f,(M.t .- axes(Legendre(),1)).^M.a,g)
+end
+function *(P::Legendre,M::PowerLawMatrix)
```
dlfivefifty, line 113: This seems back to front: we want to overload `*(::PowKernelPoint,  P::Legendre)` to return `Legendre() PowerLawMatrix(...)`

src/stieltjes.jl
```diff
@@ -75,4 +75,119 @@ end
...
+
+struct PowerLawMatrix{T} #<: AbstractMatrix{T} # TODO
+    a::T
+    t::T
+end
+
+############
+# METHODS
+############
+function dot(f::AbstractVector{T},K::PowKernelPoint{Float64,Float64,ChebyshevInterval{Float64}},g::AbstractVector{T}) where T
```
dlfivefifty, line 93: add spaces after `,` in argument lists

src/stieltjes.jl
```diff
@@ -75,4 +75,119 @@ end
...
+end
+function *(g::Adjoint,K::PowKernelPoint{<:Any,<:Any,<:ChebyshevInterval},f::AbstractVector)
+    return dot(g',K,f)
+end
+
+############
+# EVALUATE RECURRENCE
+############
+# experimental pointwise power law integral of Legendre product by recurrence
+function pointwisedot(f::AbstractVector{T},g::AbstractVector{T},a::Real,t::Real) where T
```
dlfivefifty, line 124: I'm a bit confused here. What is `coeff`? Can he adaptivity be done inside here? That is, let `f` and `g` be infinite.

dlfivefifty (changes requested):

src/stieltjes.jl
TSGut: With the way this currently stands (which I expect will change substantially) adaptivity can only work outside of that function by supplying larger pieces of f and g, i.e. overload dot(f,K,g) to do the adaptivity part when f and g are infinite. 

Is there a more natural way to have an adaptive scalar? That is, since dot(f,K,g) is the integral and returns a scalar value, what exactly would it mean to give adaptive output here? Is there a comparable example somewhere that I can have a look at?

TSGut:
Alright, with some minor tweaks to the AbstractCachedMatrix layout, this operator is now adaptive while still supporting BigFloat values. Next up is re-implementing the dot product stuff to work with this version!

src/stieltjes.jl
```diff
@@ -75,4 +75,165 @@ end
...
+
+    function PowerLawIntegral{T, PP}(P::PP, a::T, t::T) where {T, PP<:AbstractQuasiMatrix}
+        new{T, PP}(P,a,t, pointwisecoeffmatrixdense(a,t,10),(10,10))
+    end
+end
+PowerLawIntegral(P::AbstractQuasiMatrix, a::T, t::T) where T = PowerLawIntegral{T,typeof(P)}(P,a,t)
+size(K::PowerLawIntegral) = (∞,∞) # potential to add maximum size of operator
+
+# data filling
+function _legendrepowerlaw_fill_data!(K::PowerLawIntegral, inds)
```
dlfivefifty, line 133: Why this extra function?

TSGut: Just a leftover from the original structure I started with and can safely be removed.

TSGut:
@dlfivefifty I think this mostly does what I want it to now - anything else needed to make it work with Lanczos?

src/stieltjes.jl
```diff
@@ -75,4 +74,173 @@ end
...
     T[kr,:] * Diagonal(Vcat(2*convert(V,π)*c*log(c), 2c*D.diag.args[2]))
-end
\ No newline at end of file
+end
+
+#################################################
+# ∫f(x)g(x)(t-x)^a dx evaluation where f and g given in coefficients
+#################################################
+# recognize structure of W = P' * ((t .- x).^a .* P)
+const PowKernelIntegralPoint{T, V, D, F} = QuasiArrays.ApplyQuasiArray{T, 2, typeof(*), Tuple{QuasiArrays.QuasiAdjoint{T, Legendre{Float64}}, QuasiArrays.BroadcastQuasiArray{T, 2, typeof(*), Tuple{QuasiArrays.BroadcastQuasiArray{T, 1, typeof(^), Tuple{ContinuumArrays.AffineQuasiVector{T, V, Inclusion{V, D}, T}, F}}, Legendre{Float64}}}}} # method only works for Legendre for now, so require Legendre OPs.
```
dlfivefifty, line 83: This is weird. What we want is `(t .- x) .^ a .* P` to return `P * PowKernelMatrix(...)` so that `P' * ((t .- x) .^ a .* P) == P'P * PowKernelMatrix(...) = Diagonal(...) * PowKernelMatrix(...)`

src/stieltjes.jl
```diff
@@ -75,4 +74,173 @@ end
...
+end
+function *(g::Adjoint, K::PowKernelIntegralPoint, f::AbstractVector)
+    return dot(g',K,f)
+end
+
+####
+# cached operator implementation
+####
+# Constructors support BigFloat and it's recommended to use them for high orders.
+mutable struct PowerLawIntegral{T, PP<:AbstractQuasiMatrix} <: AbstractCachedMatrix{T}
```
dlfivefifty, line 116: This is a bad name: there is no `Integral`. It's literally just multiplication, that is, this represents the exact same matrix as `(t*I - X)^a` if `X` is the jacobi matrix.

src/stieltjes.jl
```diff
@@ -75,4 +74,173 @@ end
...
+function dot(f::AbstractVector{T}, K::PowKernelIntegralPoint, g::AbstractVector{T}) where T
+    # generate operator
+    a = K.args[2].args[1].args[2]
+    t = K.args[2].args[1].args[1][0.]
+    PP = K.args[2].args[2]
+    K = PowerLawIntegral(PP,BigFloat("$a"),BigFloat("$t"))
+    # check if infinite
+    (lf, lg) = (length(f),length(g))
+    maxl = maximum((lf, lg))
+    (maxl<∞) && return dot(pad(f,maxl),K[1:maxl,1:maxl],pad(g,maxl))
```
dlfivefifty, line 98: `f` and `g` should always be infinite as otherwise `K * g` doesn't make sense.

src/stieltjes.jl
dlfivefifty: To make this work we need to capture the broadcasted machinery. Something like this:
```julia
julia> const RPowKernelPointBroadcast{F<:Real} = Base.Broadcast.Broadcasted{<:Any,<:Any,typeof(^),<:Tuple{AffineQuasiVector,F}}
Base.Broadcast.Broadcasted{var"#s59",var"#s58",typeof(^),var"#s57"} where var"#s57"<:Tuple{AffineQuasiVector,F} where var"#s58" where var"#s59" where F<:Real

julia> Base.broadcasted(::LazyQuasiArrayStyle, ::typeof(*), K::RPowKernelPointBroadcast, P::Legendre) = "foo" # instead return P * PowKernelMatrix(...)

julia> (t .- x).^a .* P
"foo"
```

test/test_stieltjes.jl
```diff
@@ -68,3 +68,100 @@ end
...
+        t = 1.0000000001
+        Acached = PowerLawIntegral(Legendre(),a,t)
+        @test size(Acached) == (∞,∞)
+        @test Acached[1:20,1:20] == pointwisecoeffmatrixdense(a,t,20)
+    end
+    @testset "PowKernelPoint dot evaluation finite" begin
+        # basis
+        P = Legendre()
+        x = axes(P,1)
+        # set 1, same length coefficient vectors
```
dlfivefifty, line 88: Better to do  `@testset "set 1, same length coefficient vectors"` since then its easier to track down failures

dlfivefifty (changes requested):

src/stieltjes.jl
TSGut: Yeah, I knew something was still weird about the way this was set up. I actually have all of what you said there already working, just wasn't sure what structure we want. This should help me fix it with the next commit, thanks!

src/stieltjes.jl
TSGut: This isn't quite true as it's currently implemented but we can easily make it so that it is!

Put simply, the matrix that we are generating here is already P'P * PowKernelMatrix(...) in your terminology, so it isn't PowKernelMatrix(...) itself and it gives a different result from (t*I - X)^a (it's that but scaled by Legendre orthogonality coefficients). So that's why I called it PowerLawIntegral, because it does already contain the knowledge that it's there for integration.

We have a lot of ways to implement this differently but the recurrence method I derived much more naturally produces P'P * PowKernelMatrix(...). If you want PowKernelMatrix itself then the most simple way to get it is to compute PowerLawIntegral as it currently stands and then remove the scaling by multiplying with a diagonal matrix.

This seems in line with your other comment about wanting `P' * ((t .- x) .^ a .* P) == P'P * PowKernelMatrix(...)`, so I guess I'll just change it to that?

TSGut:
This version now satisfies what you described in https://github.com/JuliaApproximation/ClassicalOrthogonalPolynomials.jl/pull/3#discussion_r568150696

If we want to just do repeated fast integration, it's faster to just use PowerLawIntegral, which directly computes and caches (P'P)*PowerLawMatrix(...). But for simple applications PowerLawMatrix itself is fine and does exactly what (t*I-jacobimatrix(P))^a would do.

src/stieltjes.jl
dlfivefifty: Does this just mean we want to focus on `Normalized(Legendre()`? So that `P'P == I`?

TSGut: If P'P=I, then (t*I - X)^a is symmetric and indeed PowerLawIntegral = PowerLawMatrix. I haven't done it but am fairly certain I can modify the recurrence to be valid for normalized Legendre instead. Then there would be no need to have two types of operators here. It's up to - if you'd prefer that I can do it.

dlfivefifty: I think it will be a lot cleaner.

We may wish to improve `Normalized(Legendre())` to use explicit formulae for the Jacobi matrix instead of the generic fall back (which is to symmetrize the Jacobi operator)

TSGut: That's just n/sqrt((4*n^2-1)) on the sub and super diagonal, right? Should I overload it for normalized Legendre in normalized.jl?

dlfivefifty: Since normalized.jl is loaded first it should be in jacobi.jl. (Or if you feel like it you could move out the `Legendre()` code to a new file, legendre.jl). Something like this would probably be the simplest:
```julia
function jacobimatrix(::Normalized{<:Any,<:Legendre})
   b = ...#  super-diagonal as lazy BroadcastVector. Note first entry is ignored
    Symmetric(_BandedMatrix(b', (-1,1)))
end
recurrencecoefficients(:: Normalized{<:Any,<:Legendre}) = ...
```
(Unfortunately we can't use `SymTridiagonal(Zeros(∞), b)` since `SymTridiagonal` requires both vectors to be the same type. Down the line I may make my own more general version of `SymTridiagonal`.)

dlfivefifty: > That's just n/sqrt(4*n^2-1) on the sub and super diagonal, right? 

No idea but easy to check for yourself by making sure you match current behaviour: 
```julia
julia> jacobimatrix(Normalized(Legendre()))
∞×∞ Symmetric{Float64,BandedMatrices.BandedMatrix{Float64,LazyArrays.ApplyArray{Float64,2,typeof(vcat),Tuple{Adjoint{Float64,FillArrays.Zeros{Float64,1,Tuple{InfiniteArrays.OneToInf{Int64}}}},Adjoint{Float64,LazyArrays.BroadcastArray{Float64,1,typeof(/),Tuple{LazyArrays.BroadcastArray{Float64,1,typeof(*),Tuple{LazyArrays.BroadcastArray{Float64,1,typeof(/),Tuple{InfiniteArrays.InfStepRange{Float64,Float64},InfiniteArrays.InfStepRange{Int64,Int64}}},ClassicalOrthogonalPolynomials.NormalizationConstant{Float64,Legendre{Float64}}}},SubArray{Float64,1,ClassicalOrthogonalPolynomials.NormalizationConstant{Float64,Legendre{Float64}},Tuple{InfiniteArrays.InfUnitRange{Int64}},false}}}}}},InfiniteArrays.OneToInf{Int64}}} with indices OneToInf()×OneToInf():
 0.0      0.57735    ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        …  
 0.57735  0.0       0.516398   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           
  ⋅       0.516398  0.0       0.507093   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           
  ⋅        ⋅        0.507093  0.0       0.503953   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           
  ⋅        ⋅         ⋅        0.503953  0.0       0.502519   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           
  ⋅        ⋅         ⋅         ⋅        0.502519  0.0       0.501745   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        …  
  ⋅        ⋅         ⋅         ⋅         ⋅        0.501745  0.0       0.50128    ⋅         ⋅         ⋅         ⋅         ⋅           
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅        0.50128   0.0       0.500979   ⋅         ⋅         ⋅         ⋅           
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.500979  0.0       0.500773   ⋅         ⋅         ⋅           
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.500773  0.0       0.500626   ⋅         ⋅           
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.500626  0.0       0.500517   ⋅        …  
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.500517  0.0       0.500435     
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.500435  0.0          
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.50037      
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           
 ⋮                                                ⋮                                                 ⋮                             ⋱  
```

TSGut: Okay, yes that checks out. I'll add that in a first step and then modify the PowerLawMatrix stuff to be for normalized Legendre polynomials instead.

test/test_normalized.jl
```diff
@@ -3,14 +3,16 @@ import ClassicalOrthogonalPolynomials: NormalizationConstant, NormalizedBasisLay
...
+
 @testset "Normalized" begin
     @testset "Legendre" begin
         P = Legendre()
         Q = Normalized(P)
 
         @testset "Basic" begin
             @test MemoryLayout(Q) isa NormalizedBasisLayout
-            @test @inferred(Q\Q) ≡ Eye(∞)
+            @test_broken @inferred(Q\Q) ≡ Eye(∞)
```
TSGut, line 15: This line throws an error for me. I'll have to check if this is somehow caused on my end or if this is happening on master branch too. Everything else seems fine with this.

dlfivefifty: What's the error?

TSGut: I'm getting

```julia
julia> P = Legendre()
Legendre{Float64}

julia> Q = Normalized(P)
Normalized(Legendre{Float64})

julia> Q\Q
∞×∞ LinearAlgebra.Diagonal{Float64, Ones{Float64, 1, Tuple{InfiniteArrays.OneToInf{Int64}}}} with indices OneToInf()×OneToInf()

julia> Eye(∞)
∞×∞ LinearAlgebra.Diagonal{Float64, Ones{Float64, 1, Tuple{InfiniteArrays.OneToInf{Int64}}}} with indices OneToInf()×OneToInf()

julia> @inferred(Q\Q) ≡ Eye(∞)
ERROR: LoadError: return type LinearAlgebra.Diagonal{Float64, Ones{Float64, 1, Tuple{InfiniteArrays.OneToInf{Int64}}}} does not match inferred return type Any
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] top-level scope
   @ ~/Documents/Projects/ClassicalOrthogonalPolynomials.jl/test/test_normalized.jl:11
```

dlfivefifty: hmm that's also broken on mine...

dlfivefifty: ohhh... is the error being thrown in the REPL or when running as a test? 

The REPL isn't as good for type inference... but we can always just remove the `@inferred`

TSGut: Yes, this was in REPL. In that case the normalized Legendre stuff based on explicit formulae seems to work fine then.

src/stieltjes.jl
TSGut: @dlfivefifty The recurrence seems to be unstable when used to directly generate the normalized Legendre version, it degenerates after only a couple dozen orders, whereas it's perfectly stable into thousands and higher if we do Diagonal() * unnormalized(). Not exactly sure why that is, but it doesn't look like there's an obvious way around it.

dlfivefifty: that's very weird..

TSGut: I thought so too. I guess I'll keep looking for a bug that could be causing this but it certainly would be a strange bug to be accurate to order 20-30 and then become unstable when I'm basically just modifying the recurrence with the normalizations.

TSGut: Okay I figured it out but basically by luck. There was a strange interaction caused by a call to BigFloat in a wrong place. Commit with normalized version coming up.

TSGut:
Alright, this all works now:

```julia
# basis
P = Normalized(Legendre())
x = axes(P,1)
# some functions
f = P \ exp.(x.^2)
g = P \ (sin.(x).*exp.(x.^(2)))
# some parameters for (t-x)^a
a = BigFloat("1.23")
t = BigFloat("1.00001")
# define powerlaw multiplication
W = (t.-x).^a
# check if it can compute the integral correctly
@test g'*(P'*(W*P)*f) ≈ -2.656108697646584 # Mathematica
```

dlfivefifty:
We want to also support `a < 0`. Can you add tests for this?

TSGut:
As long as a>-1 and t>1 this approach will yield correct results, I added a few tests for negative a. There is also a variation that should work for t \in (0,1) if that's important to us. Currently only t>1 supported.

dlfivefifty:
For -1 < t < 1, do you mean with `|t-x|^a`? That doesn't seem useful at the moment but I'd recommend adding the code now as you may forget in the future when its needed.

TSGut:
> For -1 < t < 1, do you mean with `|t-x|^a`?

Yes, since otherwise the result is complex. Actually, the implementation as is already can do -1 < t < 1 if a is an integer but when a is integer it's much better to just use the Jacobi matrix.

dlfivefifty:
> Actually, the implementation as is already can do -1 < t < 1 if a is an integer but when a is integer it's much better to just use the Jacobi matrix.

How can the Jacobi matrix deal with `|t -  x|`? Or did you mean an even integer?

What about `t > 1` and `a < -1`? (For some reason I thought this might be useful but I can't remember why...)

TSGut:
Yes, what I said was probably a bit unclear. What I meant was that for (t-x)^a, not |t-x|^a, we can use the Jacobi matrix in the case where a integer and that doesn't care about whether t is in (-1,1) or >1. Note that this only really makes sense for a is integer, since otherwise we get complex numbers. For the |t-x|^a and t in (-1,1), which would result in real numbers, we would need to implement something that would look basically like our equilibrium measures power law thing.

Now as I said before for (t-x)^a without abs() and in the case where t>1, a>-1, the method works as is.

To answer your question about (t-x)^a without abs(), when a<-1....it's complicated. Basically, consider the following:
```julia
# basis
P = Normalized(Legendre())
x = axes(P,1)

# define some situation where a<-1
a = -1.7
t = 1.2
K = PowerLawMatrix(P,BigFloat("$a"),BigFloat("$t"))

# define function and expected result
gvec = P \ sin.(x)
res(x) = sin(x)/(1.2-x)^1.7

# compute (t-x)^a*sin(x) via powerlawmatrix
u = (P[:,1:100]*(K*gvec)[1:100])

# check result
y = rand(1)[1]
@test u[y] ≈ res(y)
```

When I run this and other examples like it, to my surprise, the tests pass just fine. So yes, this approach does work when a<-1 but right now I cannot tell you why. 🤣 My proof of this method's underlying recurrence requires us to be able to integrate \int_{-1}^1 (t-x)^a f_n g_m P_n(x) P_m(x) dx. Now when a<-1 I was under the impression that we lose integrability and thus the whole thing doesn't work. Apparently that's not the case, the recurrence approach keeps working for a<-1.

dlfivefifty:
When `t > 1` then you don't lose integrability

I assume you mean the standard approach works better for _positive_ integers. Perhaps for negative integers this is better. (note for ` a = -1` this becomes just the Cauchy transform) We can in any case special-case integers if needed...

TSGut:
Yes, positive integers. It actually can't do negative integers because the coefficients include terms like 1/((a+1)*(a+2)...). Maybe that can be salvaged somehow, I haven't thought about negative integer a much.

And yes of course, there you go. I must have been thinking about the power law kernel case too much, t>1 and a<-1 is integrable so the thing works. Nice. :)

dlfivefifty:
note negative integers can be applied in `O(|a|*n)` by just repeatedly applying `(X - a*I) \ v`

TSGut:
> note negative integers can be applied in `O(|a|*n)` by just repeatedly applying `(t*I-X) \ v`

It sounds like it would probably make sense to treat positive and negative integers as separate special cases then and just use repeated `(t*I-X) \ v` or a simple `(t*I-X)^a*v`. I doubt the recurrence approach will match these in speed, even if we do end up implementing an approximately banded version (right now it's doing it dense).

TSGut:
See https://github.com/JuliaApproximation/ClassicalOrthogonalPolynomials.jl/pull/34 for continuation of this.

PR #34, opened 2021-04-06
Pointwise dot() for powerlaw Legendre

TSGut:
Reviving this in light of https://github.com/JuliaApproximation/SemiclassicalOrthogonalPolynomials.jl/pull/36#issuecomment-809214144

The old fork that contained this code got so out of date with the main repo's file structure that manually merging it was a bit of a headache. So I closed https://github.com/JuliaApproximation/ClassicalOrthogonalPolynomials.jl/pull/3 and am instead continuing the work here, having manually moved over the relevant code to the appropriate sections.

We should set out some more specific goals for this so I can work towards getting it ready for merging.


dlfivefifty:
I fixed the multiplication syntax. At some point I think we need to rethink the design so it works with nested broadcasting; due to the difference between `Broadcasted` and `BroadcastQuasiArray`. I think the right thing to do is I have `t .- x` return a special type `ConvKernel` that is not type-aliased to a lazy quasi array, and similarly the other kenels will be types. 

What's left is to get Lanczos to correctly use this. The issue is that `Clenshaw` is mentioned explicitly instead of the more general `P \ (w .* P)` which would pick up this special case correctly.

TSGut:
Can you give the full input that leads to said error so I can go looking for the bug? What is t in your example?

dlfivefifty:
```julia
julia> P = Normalized(Legendre());

julia> t = 2; D = LanczosData((t .- x).^0.5, P);  resizedata!(D,100)
ERROR: DomainError with -1.052026918850564:
sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
Stacktrace:
 [1] throw_complex_domainerror(f::Symbol, x::Float64)
   @ Base.Math ~/Projects/julia-1.6/base/math.jl:33
 [2] sqrt
   @ ./math.jl:582 [inlined]
 [3] lanczos!(Ns::UnitRange{Int64}, X::LazyBandedMatrices.SymTridiagonal{Float64, FillArrays.Zeros{Float64, 1, Tuple{InfiniteArrays.OneToInf{Int64}}}, LazyArrays.BroadcastVector{Float64, typeof(sqrt), Tuple{LazyArrays.BroadcastVector{Float64, typeof(*), Tuple{LazyArrays.BroadcastVector{Float64, typeof(/), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Int64, Int64}}}, LazyArrays.BroadcastVector{Float64, typeof(/), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Int64, Int64}}}}}}}}, W::PowerLawMatrix{Float64, Normalized{Float64, Legendre{Float64}, LazyArrays.Accumulate{Float64, 1, typeof(*), Vector{Float64}, AbstractVector{Float64}}}}, γ::LazyArrays.CachedArray{Float64, 1, Vector{Float64}, FillArrays.Zeros{Float64, 1, Tuple{InfiniteArrays.OneToInf{Int64}}}}, β::LazyArrays.CachedArray{Float64, 1, Vector{Float64}, FillArrays.Zeros{Float64, 1, Tuple{InfiniteArrays.OneToInf{Int64}}}}, R::UpperTriangular{Float64, LazyArrays.CachedArray{Float64, 2, Matrix{Float64}, FillArrays.Zeros{Float64, 2, Tuple{InfiniteArrays.OneToInf{Int64}, InfiniteArrays.OneToInf{Int64}}}}})
   @ ClassicalOrthogonalPolynomials ~/Projects/ClassicalOrthogonalPolynomials.jl/src/lanczos.jl:19
 [4] resizedata!(L::LanczosData{Float64}, N::Int64)
   @ ClassicalOrthogonalPolynomials ~/Projects/ClassicalOrthogonalPolynomials.jl/src/lanczos.jl:61
 [5] top-level scope
   @ REPL[131]:1
```

dlfivefifty:
Sorry, just pushed the change that gets Lanczos to use PowKernelMatrix

TSGut:
Okay this looks like it's a stability thing to me. Looking at the operator for those parameters, the values explode for moderately high orders. Perhaps this is behavior for `t` away from 1 only but yet to confirm. Using BigFloats can overcome the problem but I'll have a look at whether there is a cleaner way.

dlfivefifty:
A Q&D solution would be to switch to Clenshaw when ` t ` is too large, though may depend on the length

TSGut:
Another quick and dirty solution, although less quick than what you mentioned, that would likely fix this would be to use the knowledge that the operator is approximately banded and consider any elements away from this approximate band to be zero after a certain cut-off. 

I think this is being caused by the off band entries decaying too fast when t is away from `1` (at least confirmed by BigFloat testing) and the recurrence using all those almost 0 terms with 64 bit precision leads to blowup. The upside of this approach would be that even if that turns out not to be the main cause, it will lead to faster code regardless.

TSGut:
Okay, I think I know how to fix this. I'll see how quickly I can get it done.

TSGut:
Okay so this should now (hopefully) be significantly improved. Had to change a few major things to stop it from being unstable on the diagonal which rippled through the operator. Still recommend BigFloat `a` and `t` for very high orders but it isn't usually required anymore and should be fine for most things judging from my testing.

There is at the very least some clean up work left for me to do but what you were trying to do with this may perhaps work now (some tests in test_lanczos.jl are still failing but I haven't had a look at those, I figured you were maybe not done with it yet?). I guess give it a go and let me know if there are any new or recurring old bugs.

PS Integer `a` is not supported by this method as is anymore. I don't think it's a priority but it shouldn't be too hard to make a special case version that catches that.

src/stieltjes.jl
```diff
@@ -185,3 +183,199 @@ for Op in (:Hilbert, :StieltjesPoint, :LogKernel, :PowKernel)
...
+function gennormalizedpower(a::T, t::T, ℓ::Int) where T <: Real
+    # initialization
+    ℓ = ℓ+3
+    coeff = zeros(T,ℓ,ℓ)
+    # construct first row via stable backwards recurrence
+    first = powerleg_backwardsfirstrow(a,t,2*ℓ+1)
+    coeff[1,:] = first[1:ℓ]
+    # contruct second row via normalized product Legendre decomposition
+    @inbounds for q = 1:ℓ-1
+        productcfs::Vector{T} = zeros(T,2*ℓ+1)
```
dlfivefifty, line 318: Why `::Vector{T}`?

TSGut: Not needed anymore, think that was a leftover from originally testing the decomposition.

dlfivefifty:
This doesn't seem to actually have a special `dot` implemented. Which means `lanczos` is very slow as its resorting to a default entry-wise implementation.

TSGut:
Should be able to get that working, do you have the Lanczos implementation ready the way you want it so I can have a look at what it's calling?

TSGut:
Okay I had a bit of a look at the Lanczos code now. Two things I noticed wrt performance of the PowerLawMatrix use: 

First, `resizedata!(L::LanczosData, N)` should be resizing the PowerLawMatrix `W` as well, as otherwise a lot of computation is wasted continuously resizing it one by one in the loop. Just running e.g. `resizedata!(L.W, (N,N))` in there as well should fix that time sink.

Second, for a reason that I'm not currently sure I understand, the infinite dot product that is repeatedly called in `lanczos!()` is really slow. I'll look into that but basically even simple things like instead running the finite

```julia
ll = max(length(paddeddata(v)),length(paddeddata(p1)))
dot(v[1:ll],W[1:ll,1:ll],p1[1:ll])
```
which gives the same result in each step leads to *major* performance improvements. I was under the impression that running `dot(v,W,q)` does basically this when W is infinite by infinite and `v` and `p1` are padded infinite vectors as is the case here so I'm not sure why this leads to such big performance gains? With these changes I am computing N=1000 LanczosData for the test example in 1.7 seconds, which I'd say isn't terrible considering the matrix and everything are computed densely.

I'll be looking into a specialized dot product in case we want to go to significantly higher orders than this as well.

dlfivefifty:
The issue is that in this case `dot(x,A,y)` lowers to, essentially, `dot(x,A*y)` and in this case the `A*y` is not being materialized:
```julia
julia> W*g0
(ℵ₀×ℵ₀ PowerLawMatrix{Float64, Normalized{Float64, Legendre{Float64}, LazyArrays.Accumulate{Float64, 1, typeof(*), Vector{Float64}, AbstractVector{Float64}}}} with indices OneToInf()×OneToInf()) * (ℵ₀-element LazyArrays.CachedArray{Float64, 1, Vector{Float64}, FillArrays.Zeros{Float64, 1, Tuple{InfiniteArrays.OneToInf{Int64}}}} with indices OneToInf()) with indices OneToInf():
 0.29235326262848566
 0.8486280868587607
 0.2995078694686121
 0.09545192400424624
 0.046096289203749814
 0.021838001020098422
 0.010112428144958219
 0.004718431389025629
 0.0022140477566005177
 0.0010428170625925374
 0.0004926863673495668
 0.00023337333978329177
 ⋮
```

TSGut:
I see that makes sense. So is there a reason that the three dot product `dot(v,M,w)` with v and w PaddedVectors isn't overloaded to not do that - would that introduce issues elsewhere? Seems to me that would in general be a good idea.

TSGut:
Alright, well in that case adding something more specific like

```julia
function dot(v, W::PowerLawMatrix, q)
    vpad, qpad = paddeddata(v), paddeddata(q)
    vl, ql = length(vpad), length(qpad)
    return dot(vpad,W[1:vl,1:ql],qpad)
end
``` 
will make LanczosData compute in more sensible times for now.

Maybe there is a nice way to specialize the dot product but at the moment I can't think of a way that doesn't lose the advantage of having the cached operator for repeat use. I'll think about it.

TSGut:
@dlfivefifty Can you check whether this change now fixes the speed issue you were seeing? I'm getting

```julia
julia> P = Normalized(Legendre());
julia> x = axes(P,1);
julia> @time D = ClassicalOrthogonalPolynomials.LanczosData((1.001 .- x).^0.61, P)
  0.000958 seconds (2.58 k allocations: 285.844 KiB)
julia> @time ClassicalOrthogonalPolynomials.resizedata!(D,1000)
  1.849313 seconds (26.38 k allocations: 5.030 GiB, 11.27% gc time)
```

PR #60, opened 2021-08-03
WIP Support differing intervals in Hilbert

dlfivefifty:


dlfivefifty:
@putianyi889 This might be a good example of implementing an `AbstractCachedMatrix`

@tsgut this seems _very_ similar to the power law matrices. I'm wondering if there is a simple recurrence that the entries satisfy that we can use.

TSGut:
A recurrence for the entries of Hilbert acting on weighted ChebyshevU?

dlfivefifty:
Yes but off the interval. 

If you look at the code you can see the first column is just mapped Chebyshev T coefficients of ` ξ = x- sqrt(x^2 - 1) ` . Other columns are coefficients of ` ξ^k `. Right now I use operator Clenshaw to build column by column, it’s ok speed wise (less than a second) but far from optimal, and will degenerate as the intervals coalesce

TSGut:
Okay, cool. I'll have a look for one.

dlfivefifty:
The other option is "operator-valued Olver's algorithm", where we run Olver's algorithm with the Jacobi operator and get all the columns at once. This would work for any "off" Hilbert operator but will take some thought on how to implement efficiently.

PR #68, opened 2021-08-27
Log kernel point evaluation for weighted ChebyshevU

ioannisPApapadopoulos:
Implementation of (real) point evaluation of the log kernel applied to weighted ChebyshevU polynomials. Currently only works for reals, I am not quite sure why the complex case does not work. It does not enter the * function (as if `const LogKernelPoint` was not implemented at all).


dlfivefifty (approved):

dlfivefifty:
Code cov misses are due to macros, so fine

PR #69, opened 2021-09-01
Support Mapped-Weighted OPs

dlfivefifty:


dlfivefifty:
Good news: I got the two-interval Hilbert inversion working. It's slow at the moment (10s!!) but that should just take some lazy array magic to avoid calling `getindex` directly to improve considerably, I'm sure without too much effort I can get something like 0.05s.

Combined with the log kernel point code this should be everything we need to get the two-interval equilibrium measure code working.

dlfivefifty:
OK it's now less than a second.

There's still room for improvement by adding a special case for `copyto!` with a `BlockBroadcastArray(hvcat, ...)`. Though this is a bit tricky to get right: I want to efficiently populate a `BlockBandedMatrix` with banded matrices. So we would need something like:
```julia
dest = BlockBandedMatrix(...)
k,j = 1,1 # block-indices
V = view(dest, getindex.(Block.(1:N),k), getindex.(Block.(1:N),j))
MemoryLayout(V) isa BandedColumns{StridedLayout}
```
There's an added challenge that we need to redesign `dest` so that it has dummy entries, otherwise it isn't in a `BandedColumns` format. 

Since performance isn't so critical right now maybe we just live with it for now...though this is useful for other problems e.g. vector-ODEs.

ioannisPApapadopoulos:
Thanks @dlfivefifty! There's currently some package issues on this branch/pr. Running `using ClassicalOrthogonalPolynomials` gives

```
WARNING: could not import LazyArrays.PaddedArray into ClassicalOrthogonalPolynomials
WARNING: could not import InfiniteLinearAlgebra.chop! into ClassicalOrthogonalPolynomials
```

which does not appear on the main branch. That causes the tests to fail since `chop!` is not imported.

dlfivefifty:
Getting things tagged now. Though if you check the open PRs in the various packages it should be clear which ones are active

ioannisPApapadopoulos:
Brilliant, thank you!

PR #78, opened 2021-09-27
WIP: Some powerlaw Legendre dot improvements

TSGut:
As I'm going through this at the moment I will also try to update some of this code where I see an opportunity, for now it's WIP.

As a start I replaced the Miller with a forward recurrence - not sure why I opted for Miller in this instance as it allows for a backdoor implicit t dependence. We don't need it as forward works fine, so we shouldn't use it. As far as I can tell this change restores full `t` independence of the method at least in terms of the algorithm - whether a stability analysis would end up with some `t` dependence somewhere I wouldn't want to bet but it still works as is for things like `t=1.000001` so if that is a thing it at least isn't a big deal in practice.

On computational cost, this implementation of the operator generation appears to be O(n^2) in order to compute an `n x n` block 
from scratch according to numerical experiments:

cost

To do anything useful with this method we'd of course need to consider computing `dot(f,W,g)`, so this isn't by any means all the cost there is here but it's the cost contributed by this when including it in whatever other method.


TSGut:
Going through and closing some stale PRs / finishing up ones we want, so I will close this one. 

For future reference, this one was a stability improvement for PowerLawMatrix but we decided we don't need it or if we do we'd want to redesign it. FWIW the above comment roughly shows the complexity of the approach, if we ever want to reconsider building the operator from multiple recurrences that can be a point of comparison.

PR #113, opened 2022-11-24
Laguerre Derivatives

ioannisPApapadopoulos:
@dlfivefifty Is there a reason why the LaguerreWeight axis was originally all of R rather than just the HalfLine?


dlfivefifty:
Sounds like a bug. Note DomainSets.jl has
```julia
julia> HalfLine()
0.0..Inf (closed–open) (HalfLine)
```

test/test_stieltjes.jl
```diff
@@ -141,6 +141,15 @@ end
             @test H * ChebyshevTWeight() ≡ QuasiZeros{Float64}((x,))
             @test H * ChebyshevUWeight() == π*x
             @test (H * LegendreWeight())[0.1] ≈ log((0.1+1)/(1-0.1))
+            
+            x = axes(Laguerre(), 1)
+            @test (inv.(x .- x') * LaguerreWeight())[0.1] ≈ exp(-0.1)*expinti(0.1)
+            @test isapprox((inv.(x .- x') * LaguerreWeight(-0.5))[0.1],  3.317769, rtol=1e-4) # Mathematica
```
dlfivefifty, line 147: Why are you only using 4 digits from Mathematica? Here's how you can get 19 digits:

image

dlfivefifty (changes requested):

test/test_stieltjes.jl
ioannisPApapadopoulos: Ah, this works fine! Whatever I was doing before with NIntegrate was struggling to converge.

ioannisPApapadopoulos:
@dlfivefifty It's currently failing because `L \ (x .* L)` goes via` \(L::Laguerre, K::Laguerre)` which seems a bit odd..

dlfivefifty:
It's not odd because `x .* L` simplifies to `L * jacobimatrix(L)` then `L \ (L * jacobimatrix(L))` simplifies to `(L \L) * jacobimatrix(L)`. So we want `L\L` to work (at least the case the parameters are the same in which case it should return `SquareEye{T}(∞)`)

ioannisPApapadopoulos:
Yep, just clocked that! The same thing was causing `Laguerre(a+1) \ Derivative(x)*Laguerre(a)` to fail. All fixed.

dlfivefifty:
This should be split into two PRs: one in SingularIntegrals.jl

dlfivefifty (approved):

PR #120, opened 2023-01-23
Support Jacobi matrix computation via Cholesky and QR

TSGut:
After the unnecessary episode of the premature merger, here is the current status of this PR. Still a lot left to sort out but the functionality is there and works. The old PR with chat history is [here](https://github.com/JuliaApproximation/ClassicalOrthogonalPolynomials.jl/pull/63).

The big bug to fix for this PR is here: https://github.com/JuliaLinearAlgebra/InfiniteLinearAlgebra.jl/issues/123
Due to that bug I currently have to rely on a very adhoc fix for loss of positive definiteness.

This branch is intended to be used with https://github.com/JuliaApproximation/SemiclassicalOrthogonalPolynomials.jl/pull/70, but both are still WIP.


src/choleskyQR.jl
```diff
@@ -0,0 +1,141 @@
...
+end
+function cholesky_jacobimatrix(W::AbstractMatrix, P::OrthogonalPolynomial)
+    !(P isa Normalized) && error("Polynomials must be orthonormal.")
+    !(W isa Symmetric) && error("Weight modification matrix must be symmetric.")
+    bands = CholeskyJacobiBands(W, P)
+    return SymTridiagonal(bands[1,:],bands[2,:])
+end
+
+# The generated Jacobi operators are symmetric tridiagonal, so we store their data as two bands
+mutable struct CholeskyJacobiBands{T} <: AbstractCachedMatrix{T}
```
dlfivefifty, line 15: Why is there both a `data` and an `array`? Explain what the fields are.

src/choleskyQR.jl
```diff
@@ -0,0 +1,141 @@
...
+        K.data[axes(olddata)...] = olddata
+        inds = νμ:nm
+        cache_filldata!(K, inds)
+        K.datasize = nm
+    end
+    K
+end
+function cache_filldata!(J::CholeskyJacobiBands, inds::UnitRange{Int})
+    for k in inds
+        J.data[1,k] = (J.U * (J.X * (J.U \ [zeros(k-1); 1; zeros(∞)])))[k]
```
dlfivefifty, line 52: This doesn't have the right complexity: we can recover the entries in $O(1)$ operations whilst this is $O(k)$.

src/choleskyQR.jl
```diff
@@ -0,0 +1,141 @@
...
+end
+function qr_jacobimatrix(sqrtW::AbstractMatrix, P::OrthogonalPolynomial, checks::Bool = true)
+    checks && !(P isa Normalized) && error("Polynomials must be orthonormal.")
+    checks && !(sqrtW isa Symmetric) && error("Weight modification matrix must be symmetric.")
+    bands = QRJacobiBands(sqrtW,P)
+    K = SymTridiagonal(bands[1,:],bands[2,:])
+    return K
+end
+
+# The generated Jacobi operators are symmetric tridiagonal, so we store their data as two bands
```
dlfivefifty, line 86: Add a docstring. Again, why both data and array?

dlfivefifty (changes requested):

src/choleskyQR.jl
TSGut: How does one compute $U X U^{-1}$ in $O(1)$ complexity? That would be a big improvement obviously.

src/choleskyQR.jl
TSGut: I am pretty sure array goes unused and is just a leftover from a previous version now. I will remove it. I think it is self-explanatory what the other fields do but I can add a docstring anyway since it isn't a lot of work

src/choleskyQR.jl
TSGut: ah, I think I get it now. expand by all new inds instead of one by one. should be straightforward to change

dlfivefifty: Write the enty as `y' * U^{-1} e_j ` where `y = (X'U'e_k)` has only the non-zero entries `y[k:k+u+1]` where `u` is the upper-bandwidth of `U`. That is, we only need `(U \ e_j)[max(j,k):min(k+u+1,j)]` (`j = k-1:k+1` as other entries are zero)

TSGut: I think it's `y[k-1:k+u+1]` that is non-zero but I understand yeah

TSGut:
ok @dlfivefifty, this should be better. let me know if you spot anything else. Tests are passing and I am pretty much happy with it as is for now.

I think we also want to implement sqrtweighted but I would rather do that as a separate PR once this is merged

src/choleskyQR.jl
TSGut: see the other comment on this

src/choleskyQR.jl
```diff
@@ -0,0 +1,170 @@
...
+function cholesky_jacobimatrix(W::AbstractMatrix, P::OrthogonalPolynomial, checks::Bool = true)
+    checks && !(P isa Normalized) && error("Polynomials must be orthonormal.")
+    checks && !(W isa Symmetric) && error("Weight modification matrix must be symmetric.")
+    bands = CholeskyJacobiBands(W, P)
+    return SymTridiagonal(bands[1,:],bands[2,:])
+end
+
+# The generated Jacobi operators are symmetric tridiagonal, so we store their data as two bands
+mutable struct CholeskyJacobiBands{T} <: AbstractCachedMatrix{T}
+    data::Matrix{T}         # store so-far computed block of entries
```
dlfivefifty, line 26: Storage should probably be two vectors corresponding to the diagonals, not a matrix.

src/choleskyQR.jl
```diff
@@ -0,0 +1,170 @@
...
+    datasize::Int           # size of so-far computed block 
+end
+
+# Computes the initial data for the Jacobi operator bands
+function CholeskyJacobiBands(W::Symmetric{T}, P::OrthogonalPolynomial) where T
+    U = cholesky(W).U
+    X = jacobimatrix(P)
+    dat = zeros(T,2,10)
+    UX = U*X
+    dat[1,1] = (UX * (U \ [1; zeros(∞)]))[1]
```
dlfivefifty, line 38: Probably the `∞` is superfluous

dlfivefifty: Also, don't hard-code `Float64`

src/choleskyQR.jl
```diff
@@ -0,0 +1,170 @@
...
+    dat[2,1] = (UX * (U \ [zeros(1); 1; zeros(∞)]))[1]
+    @inbounds for k in 2:10
+        yk = view(UX,k,k-1:k+1)
+        dat[1,k] = dot(yk[1:2], U[k-1:k,k-1:k] \ [0; 1])
```
dlfivefifty, line 42: I'm worried about all the allocations, can you do a profile?

src/choleskyQR.jl
```diff
@@ -0,0 +1,170 @@
...
+returns the Jacobi matrix `X` associated to a quasi-matrix of polynomials
+orthogonal with respect to `w(x) w_p(x)` where `w_p(x)` is the weight of the polynomials in `P`.
+
+The resulting polynomials are orthonormal on the same domain as `P`. The supplied `P` must be normalized. Accepted inputs for `sqrtw` are the square root of the weight modification as a function or `sqrtW` as an infinite matrix representing multiplication with the function `sqrt(w)` on the basis `P`.
+
+An optional bool can be supplied, i.e. `qr_jacobimatrix(sqrtw, P, false)` to disable checks of symmetry for the weight multiplication matrix and orthonormality for the basis (use with caution).
+"""
+function qr_jacobimatrix(sqrtw::Function, P::OrthogonalPolynomial, checks::Bool = true)
+    checks && !(P isa Normalized) && error("Polynomials must be orthonormal.")
+    sqrtW = (P \ (sqrtw.(axes(P,1)) .* P)) # Compute weight multiplication via Clenshaw
```
dlfivefifty, line 97: Just call the `qr_jacobimatrix(sqrtw, P, checks=checks)`

src/choleskyQR.jl
```diff
@@ -0,0 +1,170 @@
...
+    bands = QRJacobiBands(sqrtW,P)
+    K = SymTridiagonal(bands[1,:],bands[2,:])
+    return K
+end
+function qr_jacobimatrix(sqrtW::AbstractMatrix, P::OrthogonalPolynomial, checks::Bool = true)
+    checks && !(P isa Normalized) && error("Polynomials must be orthonormal.")
+    checks && !(sqrtW isa Symmetric) && error("Weight modification matrix must be symmetric.")
+    bands = QRJacobiBands(sqrtW,P)
+    K = SymTridiagonal(bands[1,:],bands[2,:])
```
dlfivefifty, line 106: This is very worrying: I don't believe indexing `bands` this way can be fast?

dlfivefifty (changes requested):

src/choleskyQR.jl
TSGut: It is being stored as a (2,N) matrix storing the two vectors. Is there a meaningful difference between that and storing it as two separate vectors? The nice thing about this is that I only have to resize one object but I can change it if you think this matters.

src/choleskyQR.jl
```diff
@@ -0,0 +1,170 @@
...
+    U::UpperTriangular      # store upper triangular conversion matrix (needed to extend available entries)
+    X::SymTridiagonal{T}    # store the Jacobi matrix of the original basis (needed to extend available entries)
+    datasize::Int           # size of so-far computed block 
+end
+
+# Computes the initial data for the Jacobi operator bands
+function CholeskyJacobiBands(W::Symmetric{T}, P::OrthogonalPolynomial) where T
+    U = cholesky(W).U
+    X = jacobimatrix(P)
+    dat = zeros(T,2,10)
```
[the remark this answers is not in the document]
TSGut: If the question is why (2,N), then it's because we store it as two vectors in one resizeable unit. If the question is why 10,  then that is an arbitrary initialisation number for the first computation.

src/choleskyQR.jl
TSGut: good catch on the float64, will update

src/choleskyQR.jl
TSGut: yup, will take a look

src/choleskyQR.jl
TSGut: I take it you mean that you want to store it as two vectors as stated in the other comment. Should be no problem to change it

src/choleskyQR.jl
```diff
@@ -0,0 +1,170 @@
...
+returns the Jacobi matrix `X` associated to a quasi-matrix of polynomials
+orthogonal with respect to `w(x) w_p(x)` where `w_p(x)` is the weight of the polynomials in `P`.
+
+The resulting polynomials are orthonormal on the same domain as `P`. The supplied `P` must be normalized. Accepted inputs are `w` as a function or `W` as an infinite matrix representing multiplication with the function `w` on the basis `P`.
+
+An optional bool can be supplied, i.e. `cholesky_jacobimatrix(sqrtw, P, false)` to disable checks of symmetry for the weight multiplication matrix and orthonormality for the basis (use with caution).
+"""
+function cholesky_jacobimatrix(w::Function, P::OrthogonalPolynomial, checks::Bool = true)
+    checks && !(P isa Normalized) && error("Polynomials must be orthonormal.")
+    W = Symmetric(P \ (w.(axes(P,1)) .* P)) # Compute weight multiplication via Clenshaw
```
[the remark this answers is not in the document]
TSGut: done

src/choleskyQR.jl
TSGut: done

src/choleskyQR.jl
TSGut: changed to be abstractcachedvectors instead of a 2xN matrix

src/choleskyQR.jl
TSGut: ok I think this is improved now, execution time seems fast to me and allocations scale reasonably with block size I think. but if you still see issues or numbers you wouldn't expect let me know

TSGut:
@dlfivefifty: Another round of minor improvements, with band storage now styled after the band storage in SemiclassicalOPs, i.e. like in https://github.com/JuliaApproximation/SemiclassicalOrthogonalPolynomials.jl/blob/5e4d53d5d875d54a8b459839237a7c537d9d859d/src/SemiclassicalOrthogonalPolynomials.jl#L110

Also some minor typing improvements here and there. 

Not sure how to share profiler results in a sensible manner here since I mostly rely on ProfileView.jl, but here are some benchmarks on my laptop:
```julia
julia> @benchmark qr_jacobimatrix(sqrtW, P, false)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  16.051 μs …   8.576 ms  ┊ GC (min … max):  0.00% … 99.45%
 Time  (median):     17.477 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   21.229 μs ± 136.794 μs  ┊ GC (mean ± σ):  11.10% ±  1.72%

    ▄█▆▃                                                        
  ▁▅████▇▅▄▃▃▂▃▂▃▃▃▃▃▃▃▃▃▃▃▃▃▂▂▂▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  16.1 μs         Histogram: frequency by time         29.5 μs <

 Memory estimate: 28.17 KiB, allocs estimate: 296.
```

```julia
julia> @benchmark qr_jacobimatrix(sqrtW, P, false)[1000,1000]
BenchmarkTools.Trial: 3487 samples with 1 evaluation.
 Range (min … max):  1.212 ms …  10.888 ms  ┊ GC (min … max): 0.00% … 84.50%
 Time  (median):     1.328 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.430 ms ± 884.072 μs  ┊ GC (mean ± σ):  6.12% ±  8.55%

       ▃▃ ▂▁▁█▅▂▅▁                                             
  ▂▂▂▂▄███████████▅▅▅▅▅▄▃▃▄▄▄▄▃▃▃▂▃▃▂▂▂▂▂▂▂▁▂▁▁▁▂▁▁▁▁▁▁▁▁▁▁▂▂ ▃
  1.21 ms         Histogram: frequency by time        1.77 ms <

 Memory estimate: 914.77 KiB, allocs estimate: 11400.
```

Compared with Lanczos for the same problem:

```julia
julia> @benchmark jacobimatrix(LanczosPolynomial(@.(wf.(x)),Normalized(legendre(0..1))))
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  290.774 μs …  15.843 ms  ┊ GC (min … max): 0.00% … 90.28%
 Time  (median):     311.731 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   331.059 μs ± 409.251 μs  ┊ GC (mean ± σ):  3.26% ±  2.60%

      ▂▆███▅▅▄▂▂                                                 
  ▁▁▃▆███████████▇▆▅▅▅▄▄▄▃▃▃▃▃▃▃▃▂▂▃▃▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  291 μs           Histogram: frequency by time          404 μs <

 Memory estimate: 97.06 KiB, allocs estimate: 833.
```
```julia
julia> @benchmark jacobimatrix(LanczosPolynomial(@.(wf.(x)),Normalized(legendre(0..1))))[1000,1000]
BenchmarkTools.Trial: 3 samples with 1 evaluation.
 Range (min … max):  1.907 s …    3.121 s  ┊ GC (min … max): 1.39% … 1.56%
 Time  (median):     2.104 s               ┊ GC (median):    2.02%
 Time  (mean ± σ):   2.377 s ± 651.585 ms  ┊ GC (mean ± σ):  1.65% ± 0.33%

  █        █                                               █  
  █▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  1.91 s         Histogram: frequency by time         3.12 s <

 Memory estimate: 307.94 MiB, allocs estimate: 12968641.
```

src/choleskyQR.jl
dlfivefifty: you can't `resize!` a matrix but you can two vectors. Plus `SymTridiagonal` takes vectors not matrices.

TSGut:
Yes, I have been benchmarking against Lanczos and it is significantly faster. 

I think this is ready to be merged if you are happy with the changes. Then I can focus on wrapping up the PR in SemiclassicalOPs as well.

dlfivefifty:
Don't we want a `CholeskyPolynomial` to replace `LanczosPolynomial` that uses this?

Or perhaps we just need a more general type:
```julia
struct GeneralOrthogonalPolynomial{T, WW, XX} <: OrthogonalPolynomial{T}
    weight::WW
    X::XX
end
```

dlfivefifty (approved):

TSGut:
I think the simplicity of the three weight terms in the hierarchy of semiclassical Jacobi means we still want to implement that separately like I have in the PR in that repo since a lot of simplifications happen there for computing the hierarchy step by step.

But in general yes I don't see why we can't have a GeneralOrthogonalPolynomial. Raises some design philosophy questions though, since I don't know how to generically "detect" that an arbitrary supplied weight function is the square of a polynomial, so this wouldn't be able to leverage the QR variant? We always could do CholeskyOrthogonalPolynomial and QROrthogonalPolynomial separately but the latter isn't really a nice name. 😅

I think we should definitely keep the Lanzcos approach around, though. It's an extremely helpful sanity check on these things.

TSGut:
BTW, it would be nice to tag a new version of this for use in SemiclassicalOPs.

src/choleskyQR.jl
```diff
@@ -0,0 +1,167 @@
...
+cholesky_jacobimatrix(w, P)
+
+returns the Jacobi matrix `X` associated to a quasi-matrix of polynomials
+orthogonal with respect to `w(x) w_p(x)` where `w_p(x)` is the weight of the polynomials in `P`.
+
+The resulting polynomials are orthonormal on the same domain as `P`. The supplied `P` must be normalized. Accepted inputs are `w` as a function or `W` as an infinite matrix representing multiplication with the function `w` on the basis `P`.
+
+An optional bool can be supplied, i.e. `cholesky_jacobimatrix(sqrtw, P, false)` to disable checks of symmetry for the weight multiplication matrix and orthonormality for the basis (use with caution).
+"""
+function cholesky_jacobimatrix(w::Function, P::OrthogonalPolynomial, checks::Bool = true)
```
dlfivefifty, line 11: Need to remove `::OrthogonalPolyomial` as not all orthogonal polynomials are `<: OrthogonalPolynomial`, eg, any mapped polynomials

src/choleskyQR.jl
```diff
@@ -0,0 +1,167 @@
...
+    return cholesky_jacobimatrix(W, P, false)     # At this point checks already passed or were entered as false, no need to recheck
+end
+function cholesky_jacobimatrix(W::AbstractMatrix, P::OrthogonalPolynomial, checks::Bool = true)
+    checks && !(P isa Normalized) && error("Polynomials must be orthonormal.")
+    checks && !(W isa Symmetric) && error("Weight modification matrix must be symmetric.")
+    return SymTridiagonal(CholeskyJacobiBands{:dv}(W,P),CholeskyJacobiBands{:ev}(W,P))
+end
+
+# The generated Jacobi operators are symmetric tridiagonal, so we store their data in cached bands
+mutable struct CholeskyJacobiBands{dv,T} <: AbstractCachedVector{T}
```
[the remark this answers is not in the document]
TSGut: Should take no time at all, should I just open a new PR?

src/choleskyQR.jl
TSGut: Is there a sensible replacement? I guess I can just hope the user only gives it sensible input? I can try to compute P's jacobi matrix and then catch an error to return something like "Failed to compute Jacobi matrix, check the supplied Polynomials" or something?

dlfivefifty:
You could in theory just check if it is of the form `w .^ 2` but there's no reason to add that. I think the interface would be:
```julia
OrthogonalPolynomial(w::AbstractQuasiVector) = OrthogonalPolynomial(w, orthonormalpolynomial(singularities(w)))
OrthogonalPolynomial(w::AbstractQuasiVector, P::AbstractQuasiMatrix) = GeneralOrthogonalPolynomial(w, cholesky_jacobimatrix(w, P))
```
But this does have a catch: we lose reference to the `P` which we know we can convert to-from.

So perhaps a better option is something like:
```julia
# represent an Orthogonal polynomial which has a conversion operator to P
struct ConvertedOrthogonalPolynomial{T, WW, XX, PP} <: OrthogonalPolynomial{T}
    weight::WW
    X::XX
    P::PP
end

abstract type AbstractConvertedOPLayout <: AbstractOPLayout end
struct ConvertedOPLayout <: AbstractConvertedOPLayout end
# also change NormalizedOPLayout <: ConvertedOPLayout

# transform to P * U if needed for differentiation, etc.
arguments(::ApplyLayout{typeof(*)}, Q::ConvertedOrthogonalPolynomial) = Q.P, Q.X.U

# also change all the NormalizedOPLayout
@inline copy(L::Ldiv{Lay,<: AbstractConvertedOPLayout}) where Lay = copy(Ldiv{Lay,ApplyLayout{typeof(*)}}(L.A, L.B))
```

Maybe I'll do the PR for this

dlfivefifty:
> I think we should definitely keep the Lanzcos approach around, though

In general better to delete unused code: keeping it around just creates admin, and Git will always keep the code around. Though in this case we are using in the tests.

dlfivefifty:
> BTW, it would be nice to tag a new version of this for use in SemiclassicalOPs.

Let me fix it first.

src/choleskyQR.jl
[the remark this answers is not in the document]
dlfivefifty: I'll just do it

src/choleskyQR.jl
dlfivefifty: Who cares if the user doesn't give sensible input? It will just error anyways so why even check??

dlfivefifty: And who's "the user"??? Me??

TSGut:
> Maybe I'll do the PR for this

It's up to you! Just assign an issue to me if you want me to do something related to this. I assume that whoever does this new PR will also make the small changes mentioned here, i.e. change "Bands" to "Band" in the QR and Cholesky case and remove the type from the function calls? 

> But this does have a catch: we lose reference to the P which we know we can convert to-from.

Remembering this is probably better

PR #122, opened 2023-01-25
Add links to related packages

TSGut:
@dlfivefifty Do you mind if we add something like this to the Readme? Would speed up my workflow and inform people what else exists that interfaces with this.


TSGut:
Changing only the Readme file makes the Windows build fail? ... 🤦

PR #134, opened 2023-04-19
Cleanup Cholesky, add ConvertedOrthogonalPolynomial

dlfivefifty:


dlfivefifty:
Need to wait for ArrayLayouts v1.0.1 before this works

src/choleskyQR.jl
```diff
@@ -1,56 +1,77 @@
...
+    W = Symmetric(Q \ (w .* Q)) # Compute weight multiplication via Clenshaw
+    return cholesky_jacobimatrix(W, Q)
 end
-function cholesky_jacobimatrix(W::AbstractMatrix, P::OrthogonalPolynomial, checks::Bool = true)
-    checks && !(P isa Normalized) && error("Polynomials must be orthonormal.")
-    checks && !(W isa Symmetric) && error("Weight modification matrix must be symmetric.")
-    return SymTridiagonal(CholeskyJacobiBands{:dv}(W,P),CholeskyJacobiBands{:ev}(W,P))
+function cholesky_jacobimatrix(W::AbstractMatrix, Q)
+    isnormalized(Q) || error("Polynomials must be orthonormal")
+    issymmetric(W) || error("Weight modification matrix must be symmetric.")
```
TSGut, line 56: The reason I checked via `isa Symmetric` is that `issymmetric()` tended to end up in infinite loops for certain types for me (maybe it was just SymTridiagonal though which would return false here anyway). Your inclusion of it here suggests it works for you? To give an example where it gets stuck for me:

```julia
julia> T = Normalized(Jacobi(0, 1))
Normalized(Jacobi(0.0, 1.0))

julia> X = jacobimatrix(T)
ℵ₀×ℵ₀ LazyBandedMatrices.SymTridiagonal{Float64, ApplyArray{Float64, 1, typeof(vcat), Tuple{Float64, BroadcastVector{Float64, typeof(/), Tuple{Float64, BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Float64, Float64}}}}}}}, BroadcastVector{Float64, typeof(sqrt), Tuple{BroadcastVector{Float64, typeof(*), Tuple{BroadcastVector{Float64, typeof(/), Tuple{BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Float64, Float64}}}, BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Float64, Float64}}}}}, ApplyArray{Float64, 1, typeof(vcat), Tuple{Float64, BroadcastVector{Float64, typeof(/), Tuple{BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Int64, Int64}, InfiniteArrays.InfStepRange{Float64, Float64}}}, BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Float64, Float64}}}}}}}}}}}} with indices OneToInf()×OneToInf():
 0.333333  0.471405    ⋅          ⋅         ⋅         ⋅           ⋅          …  
 0.471405  0.0666667  0.489898    ⋅         ⋅         ⋅           ⋅             
  ⋅        0.489898   0.0285714  0.494872   ⋅         ⋅           ⋅             
  ⋅         ⋅         0.494872   0.015873  0.496904   ⋅           ⋅             
  ⋅         ⋅          ⋅         0.496904  0.010101  0.49793      ⋅             
  ⋅         ⋅          ⋅          ⋅        0.49793   0.00699301  0.498519    …  
  ⋅         ⋅          ⋅          ⋅         ⋅        0.498519    0.00512821     
  ⋅         ⋅          ⋅          ⋅         ⋅         ⋅          0.498888       
  ⋅         ⋅          ⋅          ⋅         ⋅         ⋅           ⋅             
 ⋮                                                   ⋮                       ⋱  

julia> issymmetric(X)
# infinite loop
```

TSGut (commented):
Nice that it succeeds to such high degree, this bods well for the annulus application once this is all ready. :)

src/choleskyQR.jl
dlfivefifty: Wait, why isn't `X` a `SymTridiagonal`??

dlfivefifty: Actually I decided to just remove the check: checking the type is unreliable and too restrictive, and checking `issymmetric` might end up in an infinite loop.

TSGut: This is what I basically ended up doing by having the bool set to `false`. :smile: I agree that the checking is annoying, if it fails it fails in relatively predictable ways anyway

Issue #121, opened 2023-01-24
Lanczos Jacobi matrices and `^`


TSGut:
Another bug related to https://github.com/JuliaApproximation/SemiclassicalOrthogonalPolynomials.jl/pull/70

Jacobi matrices obtained from Lanczos don't appear to play nice with `^`:
```julia
julia>  using ClassicalOrthogonalPolynomials

julia> P = Legendre()
Legendre()

julia> x = axes(P,1)
Inclusion(-1.0..1.0 (Chebyshev))

julia> T = LanczosPolynomial(@.(2-x),P)
LanczosPolynomial{Float64} with weight with singularities LegendreWeight{Float64}

julia> X = jacobimatrix(T)
ℵ₀×ℵ₀ SymTridiagonal{Float64} Jacobi operator from Lanczos:
 -0.166667  0.552771    ⋅           ⋅           ⋅            ⋅         ⋅    ⋅    ⋅   …  
  0.552771  0.0212121  0.519377     ⋅           ⋅            ⋅         ⋅    ⋅    ⋅      
   ⋅        0.519377   0.00679908  0.508021     ⋅            ⋅         ⋅    ⋅    ⋅      
   ⋅         ⋅         0.508021    0.00221202  0.504252      ⋅         ⋅    ⋅    ⋅      
   ⋅         ⋅          ⋅          0.504252    0.000936283  0.502645   ⋅    ⋅    ⋅      
  ⋮                                                         ⋮                        ⋱  

julia> X^1
ERROR: MethodError: Cannot `convert` an object of type LazyBandedMatrices.SymTridiagonal{Float64, ClassicalOrthogonalPolynomials.LanczosJacobiBand{Float64}, ClassicalOrthogonalPolynomials.LanczosJacobiBand{Float64}} to an object of type ApplyArray{Float64, 2, typeof(*), Tuple{LazyBandedMatrices.SymTridiagonal{Float64, ClassicalOrthogonalPolynomials.LanczosJacobiBand{Float64}, ClassicalOrthogonalPolynomials.LanczosJacobiBand{Float64}}, LazyBandedMatrices.SymTridiagonal{Float64, ClassicalOrthogonalPolynomials.LanczosJacobiBand{Float64}, ClassicalOrthogonalPolynomials.LanczosJacobiBand{Float64}}}}
Closest candidates are:
[...]

PR #135, opened 2023-04-19
Replace ∞ with ℵ₀ in Lanczos sizes

TSGut:
Fix https://github.com/JuliaApproximation/ClassicalOrthogonalPolynomials.jl/issues/121


dlfivefifty:
add tests for the previously broken behaviour

PR #138, opened 2023-05-22
Implement Q-based qrjacobimatrix() in O(n)

TSGut:
@ioannisPApapadopoulos @dlfivefifty 

This PR adds an optional `method` argument to `qr_jacobimatrix()` which allows either `:Q` or `:R` as input and will do the QR raising using the chosen matrix, both in `O(n)`. The default if no method is supplied is to use `:Q`. 

The Q method more or less matches the R method in efficiency (some efficiency is lost to make sure the resulting matrix fits into an adaptive SymTridiagonal but this is also true for the R approach) - there is more optimization that could be done in principle but I am not sure it's worth it considering we can do this:

```julia
julia> P = Normalized(legendre(0..1));

julia> x = axes(P,1);

julia> J = jacobimatrix(P);

julia> wf(x) = (1-x)^2;

julia> sqrtwf(x) = (1-x);

julia> Jchol = cholesky_jacobimatrix(wf, P);

julia> JqrQ = qr_jacobimatrix(sqrtwf, P);

julia> JqrR = qr_jacobimatrix(sqrtwf, P, :R);

julia> N = 100_000
100000

julia> @time Jchol[1:N,1:N]
  1.015375 seconds (13.59 M allocations: 988.733 MiB, 11.53% gc time)
100000×100000 BandedMatrix{Float64} with bandwidths (1, 1):
 0.25      0.193649   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅       …   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
 0.193649  0.416667  0.225374   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅        0.225374  0.458333  0.236228   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅        0.236228  0.475     0.241209   ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅        0.241209  0.483333  0.243904   ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅        0.243904  0.488095  0.245525   ⋅         ⋅         ⋅       …   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅        0.245525  0.491071  0.246576   ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.246576  0.493056  0.247296   ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.247296  0.494444  0.24781      ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
 ⋮                                                 ⋮                                                ⋱                 ⋮                        ⋮                            ⋮                       
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   0.25  0.5   0.25   ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    0.25  0.5   0.25   ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅    0.25  0.5   0.25   ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅       …   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅    0.25  0.5   0.25   ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅    0.25  0.5   0.25   ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅    0.25  0.5   0.25   ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    0.25  0.5   0.25
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    0.25  0.5

julia> @time JqrQ[1:N,1:N]
  1.347002 seconds (21.48 M allocations: 1.476 GiB, 12.26% gc time)
100000×100000 BandedMatrix{Float64} with bandwidths (1, 1):
 0.25      0.193649   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅       …   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
 0.193649  0.416667  0.225374   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅        0.225374  0.458333  0.236228   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅        0.236228  0.475     0.241209   ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅        0.241209  0.483333  0.243904   ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅        0.243904  0.488095  0.245525   ⋅         ⋅         ⋅       …   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅        0.245525  0.491071  0.246576   ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.246576  0.493056  0.247296   ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.247296  0.494444  0.24781      ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
 ⋮                                                 ⋮                                                ⋱                 ⋮                        ⋮                            ⋮                       
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   0.25  0.5   0.25   ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    0.25  0.5   0.25   ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅    0.25  0.5   0.25   ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅       …   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅    0.25  0.5   0.25   ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅    0.25  0.5   0.25   ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅    0.25  0.5   0.25   ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    0.25  0.5   0.25
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    0.25  0.5

julia> @time JqrR[1:N,1:N]
  1.179295 seconds (16.09 M allocations: 1.229 GiB, 11.27% gc time)
100000×100000 BandedMatrix{Float64} with bandwidths (1, 1):
 0.25      0.193649   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅       …   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
 0.193649  0.416667  0.225374   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅        0.225374  0.458333  0.236228   ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅        0.236228  0.475     0.241209   ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅        0.241209  0.483333  0.243904   ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅        0.243904  0.488095  0.245525   ⋅         ⋅         ⋅       …   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅        0.245525  0.491071  0.246576   ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.246576  0.493056  0.247296   ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.247296  0.494444  0.24781      ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
 ⋮                                                 ⋮                                                ⋱                 ⋮                        ⋮                            ⋮                       
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   0.25  0.5   0.25   ⋅     ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    0.25  0.5   0.25   ⋅     ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅    0.25  0.5   0.25   ⋅     ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅       …   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅    0.25  0.5   0.25   ⋅     ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅    0.25  0.5   0.25   ⋅     ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅    0.25  0.5   0.25   ⋅ 
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    0.25  0.5   0.25
  ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅         ⋅           ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    0.25  0.5
```

Once this is done I will change my PR in SemiclassicalOPs to use this approach for the hierarchy.

### DIFF
--- a/src/choleskyQR.jl
+++ b/src/choleskyQR.jl
@@ -150,12 +150,13 @@ function QRJacobiBand{:dv,:Q}(F, P::OrthogonalPolynomial{T}) where T
         # fill first entry (special case)
     M = X[1:b,1:b]
     getindex(F.factors,b,b) # pre-fill cached array
-    v = I + tril(F.factors[1:b,1:b],-1)
-    H = I - F.τ[1]*v[:,1]*v[:,1]'
+    v = [one(T);F.factors[1:b,1][2:end]]
+    H = I - F.τ[1]*v*v'
     M = H*M*H
     dv[1] = M[1,1]
         # fill second entry
-    H = I - F.τ[2]*v[:,2]*v[:,2]'
+    v = [zero(T);one(T);F.factors[1:b,2][3:end]]
+    H = I - F.τ[2]*v*v'
     K = H*M*H
     M = Matrix(X[2:b+1,2:b+1])
     M[1:end-1,1:end-1] = K[2:end,2:end]
@@ -170,11 +171,12 @@ function QRJacobiBand{:ev,:Q}(F, P::OrthogonalPolynomial{T}) where T
         # first step does not produce entries for the off-diagonal band (special case)
     M = X[1:b,1:b]
     getindex(F.factors,b,b) # pre-fill cached array
-    v = (I + tril(F.factors[1:b,1:b],-1))
-    H = I - F.τ[1]*v[:,1]*v[:,1]'
+    v = [one(T);F.factors[1:b,1][2:end]]
+    H = I - F.τ[1]*v*v'
     M = H*M*H
         # fill first off-diagonal entry
-    H = I - F.τ[2]*v[:,2]*v[:,2]'
+    v = [zero(T);one(T);F.factors[1:b,2][3:end]]
+    H = I - F.τ[2]*v*v'
     K = H*M*H
     dv[1] = K[1,2]*sign(F.R[1,1]*F.R[2,2]) # includes possible correction for sign (only needed in off-diagonal case), since the QR decomposition does not guarantee positive diagonal on R
     M = Matrix(X[2:b+1,2:b+1])
@@ -226,7 +228,7 @@ function cache_filldata!(J::QRJacobiBand{:dv,:Q,T}, inds::UnitRange{Int}) where 
     F = J.U.factors
     dv = J.data
     @inbounds for n in jj
-        v = (I + tril(F[n-1:n+b,n-1:n+b],-1))[:,2]
+        v = [zero(T);one(T);F[n-1:n+b,n][3:end]]
         H = I-τ[n]*v*v'
         K = H*M*H
         M = Matrix(X[n:n+b+1,n:n+b+1])
@@ -247,7 +249,7 @@ function cache_filldata!(J::QRJacobiBand{:ev,:Q,T}, inds::UnitRange{Int}) where 
     M, τ, F, dv = J.UX, J.U.τ, J.U.factors, J.data
     D = sign.(view(J.U.R,band(0)).*view(J.U.R,band(0))[2:end])
     @inbounds for n in jj
-        v = (I + tril(F[n:n+b+2,n:n+b+2],-1))[:,2]
+        v = [zero(T);one(T);F[n:n+b+2,n+1][3:end]]
         H = I - τ[n+1]*v*v'
         K = H*M*H
         dv[n] = K[1,2]
--- a/src/choleskyQR.jl
+++ b/src/choleskyQR.jl
@@ -243,7 +243,7 @@ function cache_filldata!(J::QRJacobiBand{:ev,:Q,T}, inds::UnitRange{Int}) where 
     b = bandwidths(J.U.factors)[2]÷2
     # pre-fill cached arrays to avoid excessive cost from expansion in loop
     X = jacobimatrix(J.P)[1:m+b+2,1:m+b+2]
-    getindex(J.U.factors,m+b,m+b)
+    resizedata!(J.U.factors,m+b,m+b)
     getindex(J.U.R,m,m)
     getindex(J.U.τ,m)
     M, τ, F, dv = J.UX, J.U.τ, J.U.factors, J.data
--- a/src/choleskyQR.jl
+++ b/src/choleskyQR.jl
@@ -244,7 +244,7 @@ function cache_filldata!(J::QRJacobiBand{:ev,:Q,T}, inds::UnitRange{Int}) where 
     # pre-fill cached arrays to avoid excessive cost from expansion in loop
     X = jacobimatrix(J.P)[1:m+b+2,1:m+b+2]
     resizedata!(J.U.factors,m+b,m+b)
-    getindex(J.U.R,m,m)
+    resizedata!(J.U.R,m,m)
     getindex(J.U.τ,m)
     M, τ, F, dv = J.UX, J.U.τ, J.U.factors, J.data
     D = sign.(view(J.U.R,band(0)).*view(J.U.R,band(0))[2:end])
--- a/src/choleskyQR.jl
+++ b/src/choleskyQR.jl
@@ -145,11 +145,11 @@ mutable struct QRJacobiData{method,T} <: AbstractMatrix{T}
 end
 
 # computes H*M*H in-place, overwriting K
-function doublehouseholderapply!(M, τ, v, w)
-    mul!(w, M, v)  # M is symmetric
-    M .= M .- τ .* v .* w'
-    mul!(w, M, v)
-    M .= M .- τ .* w .* v'
+function doublehouseholderapply!(M::AbstractMatrix{T}, τ::T, v, w) where T
+    mul!(w, M, [zero(T);one(T);v])  # M is symmetric
+    M .= M .- τ .* [zero(T);one(T);v] .* w'
+    mul!(w, M, [zero(T);one(T);v])
+    M .= M .- τ .* w .* [zero(T);one(T);v]'
 end
 
 # Computes the initial data for the Jacobi operator bands
@@ -163,10 +163,15 @@ function QRJacobiData{:Q,T}(F, P) where T
     M = Matrix(X[1:b,1:b])
     resizedata!(F.factors,b,b)
     w = Vector{T}(undef, b)
-    doublehouseholderapply!(M,F.τ[1],[one(T);F.factors[2:b,1]],w)
+    # special case for first entry double Householder product
+    v = [one(T);view(F.factors,2:b,1)]
+    mul!(w, M, v)  # M is symmetric
+    M .= M .- F.τ[1] .* v .* w'
+    mul!(w, M, v)
+    M .= M .- F.τ[1] .* w .* v'
     dv[1] = M[1,1]
         # fill second entry
-    doublehouseholderapply!(M,F.τ[2],[zero(T);one(T);F.factors[3:b,2]],w)
+    doublehouseholderapply!(M,F.τ[2],view(F.factors,3:b,2),w)
     ev[1] = M[1,2]*sign(F.R[1,1]*F.R[2,2]) # includes possible correction for sign (only needed in off-diagonal case), since the QR decomposition does not guarantee positive diagonal on R
     K = Matrix(X[2:b+1,2:b+1])
     K[1:end-1,1:end-1] .= view(M,2:b,2:b)
@@ -221,13 +226,12 @@ function _fillqrbanddata!(J::QRJacobiData{:Q,T}, inds::UnitRange{Int}) where T
     M = Matrix{T}(undef,b+3,b+3)
     @inbounds for n in jj
         dv[n] = K[1,1] # no sign correction needed on diagonal entry due to cancellation
-        doublehouseholderapply!(K,τ[n+1],[zero(T);one(T);F[n+2:n+b+2,n+1]],w)
+        doublehouseholderapply!(K,τ[n+1],view(F,n+2:n+b+2,n+1),w)
         ev[n] = K[1,2]*D[n] # contains sign correction from QR not forcing positive diagonals
         M .= view(X,n+1:n+b+3,n+1:n+b+3)
         M[1:end-1,1:end-1] .= view(K,2:b+3,2:b+3)
         K .= M
     end
-    J.UX = M
 end
 
 function _fillqrbanddata!(J::QRJacobiData{:R,T}, inds::UnitRange{Int}) where T
--- a/src/choleskyQR.jl
+++ b/src/choleskyQR.jl
@@ -144,6 +144,18 @@ mutable struct QRJacobiData{method,T} <: AbstractMatrix{T}
     datasize::Int         # size of so-far computed block 
 end
 
+@inline function inplaceHouseholder!(x::AbstractVector, τ::Number, A::AbstractVecOrMat)
+    m, n = size(A, 1), size(A, 2)
+    m == 0 && return A
+    @inbounds for j = 1:n
+        Aj, xj = view(A, 2:m, j), view(x, 2:m)
+        vAj = conj(τ)*(A[1, j] + dot(xj, Aj))
+        A[1, j] -= vAj
+        axpy!(-vAj, xj, Aj)
+    end
+    return A
+end
+
 # Computes the initial data for the Jacobi operator bands
 function QRJacobiData{:Q,T}(F, P) where T
     b = 3+bandwidths(F.R)[2]÷2
@@ -156,16 +168,16 @@ function QRJacobiData{:Q,T}(F, P) where T
     resizedata!(F.factors,b,b)
     # special case for first entry double Householder product
     v = view(F.factors,1:b,1)
-    reflectorApply!(v, F.τ[1], M)
-    reflectorApply!(v, F.τ[1], M')
+    inplaceHouseholder!(v, F.τ[1], M)
+    inplaceHouseholder!(v, F.τ[1], M')
     dv[1] = M[1,1]
         # fill second entry
     # computes H*M*H in-place, overwriting M
     v = view(F.factors,2:b,2)
-    reflectorApply!(v, F.τ[2], view(M,1,2:b))
+    inplaceHouseholder!(v, F.τ[2], view(M,1,2:b))
     M[1,2:b] .= view(M,1,2:b) # symmetric matrix, avoid recomputation
-    reflectorApply!(v, F.τ[2], view(M,2:b,2:b))
-    reflectorApply!(v, F.τ[2], view(M,2:b,2:b)')
+    inplaceHouseholder!(v, F.τ[2], view(M,2:b,2:b))
+    inplaceHouseholder!(v, F.τ[2], view(M,2:b,2:b)')
     ev[1] = M[1,2]*sign(F.R[1,1]*F.R[2,2]) # includes possible correction for sign (only needed in off-diagonal case), since the QR decomposition does not guarantee positive diagonal on R
     K = Matrix(X[2:b+1,2:b+1])
     K[1:end-1,1:end-1] .= view(M,2:b,2:b)
@@ -221,10 +233,10 @@ function _fillqrbanddata!(J::QRJacobiData{:Q,T}, inds::UnitRange{Int}) where T
         dv[n] = K[1,1] # no sign correction needed on diagonal entry due to cancellation
         # doublehouseholderapply!(K,τ[n+1],view(F,n+2:n+b+2,n+1),w)
         v = view(F,n+1:n+b+2,n+1)
-        reflectorApply!(v, τ[n+1], view(K,1,2:b+3))
+        inplaceHouseholder!(v, τ[n+1], view(K,1,2:b+3))
         M[1,2:b+3] .= view(M,1,2:b+3) # symmetric matrix, avoid recomputation
-        reflectorApply!(v, τ[n+1], view(K,2:b+3,2:b+3))
-        reflectorApply!(v, τ[n+1], view(K,2:b+3,2:b+3)')
+        inplaceHouseholder!(v, τ[n+1], view(K,2:b+3,2:b+3))
+        inplaceHouseholder!(v, τ[n+1], view(K,2:b+3,2:b+3)')
         ev[n] = K[1,2]*D[n] # contains sign correction from QR not forcing positive diagonals
         M .= view(X,n+1:n+b+3,n+1:n+b+3)
         M[1:end-1,1:end-1] .= view(K,2:b+3,2:b+3)
--- a/src/choleskyQR.jl
+++ b/src/choleskyQR.jl
@@ -144,7 +144,7 @@ mutable struct QRJacobiData{method,T} <: AbstractMatrix{T}
     datasize::Int         # size of so-far computed block 
 end
 
-# computes H*M*H in-place, overwriting K
+# computes H*M*H in-place, overwriting M
 function doublehouseholderapply!(M::AbstractMatrix{T}, τ::T, v, w) where T
     k = [zero(T);one(T);v]
     mul!(w, M, k)  # M is symmetric
--- a/src/choleskyQR.jl
+++ b/src/choleskyQR.jl
@@ -143,44 +143,42 @@ end
 # Computes the initial data for the Jacobi operator bands
 function QRJacobiBand{:dv,:Q}(sqrtW, P::OrthogonalPolynomial{T}) where T
     F = qr(sqrtW)
-    bds = bandwidths(F.R)[2]÷2
+    b = 3+bandwidths(F.R)[2]÷2
     X = jacobimatrix(P)
         # we fill 2 entries on the first run
     dv = zeros(T,2)
         # fill first entry (special case)
-    M = X[1:bds+3,1:bds+3]
-    v = I + tril(F.factors[1:bds+3,1:bds+3],-1)
-    H = I-F.τ[1]*v[:,1]*v[:,1]'
+    M = X[1:b,1:b]
+    getindex(F.factors,b,b) # pre-fill cached array
+    v = I + tril(F.factors[1:b,1:b],-1)
+    H = I - F.τ[1]*v[:,1]*v[:,1]'
     M = H*M*H
     dv[1] = M[1,1]
         # fill second entry
-    n = 2
-    v = I + tril(F.factors[n-1:n+bds+1,n-1:n+bds+1],-1)
-    H = I-F.τ[2]*v[:,2]*v[:,2]'
+    H = I - F.τ[2]*v[:,2]*v[:,2]'
     K = H*M*H
-    M = Matrix(jacobimatrix(P)[n:n+bds+2,n:n+bds+2])
+    M = Matrix(X[2:b+1,2:b+1])
     M[1:end-1,1:end-1] = K[2:end,2:end]
     dv[2] = M[1,1] # sign correction due to QR not guaranteeing positive diagonal for R not needed on diagonals since contributions cancel
     return QRJacobiBand{:dv,:Q,T}(dv, F, M, P, 2)
 end
 function QRJacobiBand{:ev,:Q}(sqrtW, P::OrthogonalPolynomial{T}) where T
     F = qr(sqrtW)
-    bds = bandwidths(F.R)[2]÷2
+    b = 3+bandwidths(F.factors)[2]÷2
     X = jacobimatrix(P)
         # we fill 1 entry on the first run
     dv = zeros(T,1)
         # first step does not produce entries for the off-diagonal band (special case)
-    M = X[1:bds+3,1:bds+3]
-    v = I + tril(F.factors[1:bds+3,1:bds+3],-1)
-    H = I-F.τ[1]*v[:,1]*v[:,1]'
+    M = X[1:b,1:b]
+    getindex(F.factors,b,b) # pre-fill cached array
+    v = (I + tril(F.factors[1:b,1:b],-1))
+    H = I - F.τ[1]*v[:,1]*v[:,1]'
     M = H*M*H
         # fill first off-diagonal entry
-    n = 2
-    v = I + tril(F.factors[n-1:n+bds+1,n-1:n+bds+1],-1)
-    H = I-F.τ[2]*v[:,2]*v[:,2]'
+    H = I - F.τ[2]*v[:,2]*v[:,2]'
     K = H*M*H
-    dv[1] = K[1,2]*sign(F.R[1,1])*sign(F.R[2,2]) # includes possible correction for sign (only needed in off-diagonal case), since the QR decomposition does not guarantee positive diagonal on R
-    M = Matrix(jacobimatrix(P)[n:n+bds+2,n:n+bds+2])
+    dv[1] = K[1,2]*sign(F.R[1,1]*F.R[2,2]) # includes possible correction for sign (only needed in off-diagonal case), since the QR decomposition does not guarantee positive diagonal on R
+    M = Matrix(X[2:b+1,2:b+1])
     M[1:end-1,1:end-1] = K[2:end,2:end]
     return QRJacobiBand{:ev,:Q,T}(dv, F, M, P, 1)
 end
@@ -218,38 +216,47 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
     K
 end
 function cache_filldata!(J::QRJacobiBand{:dv,:Q,T}, inds::UnitRange{Int}) where T
-    bds = bandwidths(J.U.R)[2]÷2
+    b = 1+bandwidths(J.U.factors)[2]÷2
     # pre-fill cached arrays to avoid excessive cost from expansion in loop
-    m = inds[end]
-    X = jacobimatrix(J.P)[1:m+bds+3,1:m+bds+3]
-    getindex(J.U.factors,m+bds+1,m+bds+1)
+    m, jj = inds[end], inds[2:end]
+    X = jacobimatrix(J.P)[1:m+b+2,1:m+b+2]
+    getindex(J.U.factors,m+b,m+b)
     getindex(J.U.τ,m)
-    @inbounds for n in inds[2:end]
-        v = I + tril(J.U.factors[n-1:n+bds+1,n-1:n+bds+1],-1)
-        H = I-J.U.τ[n]*v[:,2]*v[:,2]'
-        K = H*J.UX*H
-        J.UX = Matrix(X[n:n+bds+2,n:n+bds+2])
-        J.UX[1:end-1,1:end-1] = K[2:end,2:end]
-        J.data[n] = J.UX[1,1] # sign correction due to QR not guaranteeing positive diagonal for R not needed on diagonals since contributions cancel
+    M = J.UX
+    τ = J.U.τ
+    F = J.U.factors
+    dv = J.data
+    @inbounds for n in jj
+        v = (I + tril(F[n-1:n+b,n-1:n+b],-1))[:,2]
+        H = I-τ[n]*v*v'
+        K = H*M*H
+        M = Matrix(X[n:n+b+1,n:n+b+1])
+        M[1:end-1,1:end-1] = K[2:end,2:end]
+        dv[n] = M[1,1] # sign correction due to QR not guaranteeing positive diagonal for R not needed on diagonals since contributions cancel
     end
+    J.UX = M
+    J.data[jj] = dv[jj]
 end
 function cache_filldata!(J::QRJacobiBand{:ev,:Q,T}, inds::UnitRange{Int}) where T
-    bds = bandwidths(J.U.R)[2]÷2
+    m, jj = 1+inds[end], inds[2:end]
+    b = bandwidths(J.U.factors)[2]÷2
     # pre-fill cached arrays to avoid excessive cost from expansion in loop
-    m = inds[end]
-    X = jacobimatrix(J.P)[1:m+bds+3,1:m+bds+3]
-    getindex(J.U.factors,m+bds+1,m+bds+1)
-    getindex(J.U.R,m+1,m+1)
-    getindex(J.U.τ,m+1)
+    X = jacobimatrix(J.P)[1:m+b+2,1:m+b+2]
+    getindex(J.U.factors,m+b,m+b)
+    getindex(J.U.R,m,m)
+    getindex(J.U.τ,m)
+    M, τ, F, dv = J.UX, J.U.τ, J.U.factors, J.data
     D = sign.(view(J.U.R,band(0)).*view(J.U.R,band(0))[2:end])
-    @inbounds for n in inds[2:end]
-        v = I + tril(J.U.factors[n:n+bds+2,n:n+bds+2],-1)
-        H = I-J.U.τ[n+1]*v[:,2]*v[:,2]'
-        K = H*J.UX*H
-        J.data[n] = K[1,2]*D[n] # includes possible correction for sign (only needed in off-diagonal case), since the QR decomposition does not guarantee positive diagonal on R
-        J.UX = Matrix(X[n+1:n+bds+3,n+1:n+bds+3])
-        J.UX[1:end-1,1:end-1] = K[2:end,2:end]
+    @inbounds for n in jj
+        v = (I + tril(F[n:n+b+2,n:n+b+2],-1))[:,2]
+        H = I - τ[n+1]*v*v'
+        K = H*M*H
+        dv[n] = K[1,2]
+        M = Matrix(X[n+1:n+b+3,n+1:n+b+3])
+        M[1:end-1,1:end-1] = K[2:end,2:end]
     end
+    J.UX = M
+    J.data[jj] = dv[jj].*D[jj] # includes possible correction for sign (only needed in off-diagonal case), since the QR decomposition does not guarantee positive diagonal on R
 end
 function cache_filldata!(J::QRJacobiBand{:dv,:R,T}, inds::UnitRange{Int}) where T
     # pre-fill U and UX to prevent expensive step-by-step filling in of cached U and UX in the loop
@@ -258,9 +265,11 @@ function cache_filldata!(J::QRJacobiBand{:dv,:R,T}, inds::UnitRange{Int}) where 
     getindex(J.UX,m,m)
 
     ek = [zero(T); one(T)]
+    dv, UX, U = J.data, J.UX, J.U
     @inbounds for k in inds
-        J.data[k] = dot(view(J.UX,k,k-1:k), J.U[k-1:k,k-1:k] \ ek)
+        dv[k] = dot(view(UX,k,k-1:k), U[k-1:k,k-1:k] \ ek)
     end
+    J.data[inds] = dv[inds]
 end
 function cache_filldata!(J::QRJacobiBand{:ev,:R, T}, inds::UnitRange{Int}) where T
     # pre-fill U and UX to prevent expensive step-by-step filling in of cached U and UX in the loop
@@ -269,7 +278,9 @@ function cache_filldata!(J::QRJacobiBand{:ev,:R, T}, inds::UnitRange{Int}) where
     getindex(J.UX,m,m)
 
     ek = [zeros(T,2); one(T)]
+    dv, UX, U = J.data, J.UX, J.U
     @inbounds for k in inds
-        J.data[k] = dot(view(J.UX,k,k-1:k+1), J.U[k-1:k+1,k-1:k+1] \ ek)
+        dv[k] = dot(view(UX,k,k-1:k+1), U[k-1:k+1,k-1:k+1] \ ek)
     end
+    J.data[inds] = dv[inds]
 end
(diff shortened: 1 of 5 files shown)


TSGut:
@dlfivefifty Not sure what's going on here, I think the failing tests have nothing to do with my changes. Did something recently change further upstream in the dependencies?

dlfivefifty:
I can't decide whether to ask why the errors are so large in the second example or so small in the first 😅 Since if round-off grows like N we expect errors on the order of:
```julia
julia> eps() * 100_000
2.220446049250313e-11
```

TSGut:
I don't know. 😅 But actually I think the first is more representative of the general behavior, not the second. In the second I'm doing `x` and `1-x` in one step, which is worse than doing them one at a time. In the semiclassical hierarchy we'd be doing things like in the first example.

TSGut:
Just to back up my claim with code here is an example of raising by the other weight term in one step, at a higher point in the Jacobi family. If we do a single step we see the errors observed in the first example.

```julia
julia> P = Normalized(jacobi(3,3,0..1));

julia> x = axes(P,1);

julia> J = jacobimatrix(P);

julia> sqrtwf(x) = x;

julia> JqrQ = qr_jacobimatrix(sqrtwf, P);

julia> JqrR = qr_jacobimatrix(sqrtwf, P, :R);

julia> N = 100_000
100000
julia> Jclass = jacobimatrix(Normalized(jacobi(3,5,0..1)))
ℵ₀×ℵ₀ LazyBandedMatrices.SymTridiagonal{Float64, ApplyArray{Float64, 1, typeof(vcat), Tuple{Float64, BroadcastVector{Float64, typeof(/), Tuple{BroadcastVector{Float64, typeof(-), Tuple{BroadcastVector{Float64, typeof(/), Tuple{Float64, BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Float64, Float64}}}}}, Fill{Float64, 1, Tuple{InfiniteArrays.OneToInf{Int64}}}}}, Float64}}}}, BroadcastVector{Float64, typeof(sqrt), Tuple{BroadcastVector{Float64, typeof(*), Tuple{BroadcastVector{Float64, typeof(/), Tuple{BroadcastVector{Float64, typeof(/), Tuple{BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Float64, Float64}}}, BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Float64, Float64}}}}}, Float64}}, ApplyArray{Float64, 1, typeof(vcat), Tuple{Float64, BroadcastVector{Float64, typeof(/), Tuple{BroadcastVector{Float64, typeof(/), Tuple{BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Int64, Int64}, InfiniteArrays.InfStepRange{Float64, Float64}}}, BroadcastVector{Float64, typeof(*), Tuple{InfiniteArrays.InfStepRange{Float64, Float64}, InfiniteArrays.InfStepRange{Float64, Float64}}}}}, Float64}}}}}}}}} with indices OneToInf()×OneToInf():
 0.6      0.14771    ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        …  
 0.14771  0.566667  0.184374   ⋅         ⋅         ⋅         ⋅         ⋅           
  ⋅       0.184374  0.547619  0.203579   ⋅         ⋅         ⋅         ⋅           
  ⋅        ⋅        0.203579  0.535714  0.215229   ⋅         ⋅         ⋅           
  ⋅        ⋅         ⋅        0.215229  0.527778  0.222909   ⋅         ⋅           
  ⋅        ⋅         ⋅         ⋅        0.222909  0.522222  0.228266   ⋅        …  
  ⋅        ⋅         ⋅         ⋅         ⋅        0.228266  0.518182  0.232161     
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅        0.232161  0.515152     
  ⋅        ⋅         ⋅         ⋅         ⋅         ⋅         ⋅        0.235087     
 ⋮                                                ⋮                             ⋱  

julia> norm(JqrQ[1:N,1:N]-Jclass[1:N,1:N],2)
6.784308695781622e-14

julia> norm(JqrR[1:N,1:N]-Jclass[1:N,1:N],2)
1.619843167289299e-13
```
Same behavior in terms of norms.

TSGut:
Ok actually the constant between the Q and R method is more substantial than the simple timer was suggesting because of an implementation detail in how cached vectors are expanded.

Here are two CPU timings showing the _actual_ linear complexity in both methods with current implementation.

XraisingLegendre

I prematurely stopped optimizing the Q method because I thought it was matching R but with the difference being this big (roughly 5x the cpu time) it is worth taking another shot at bringing the cost down. I have a few implementation ideas that should do it based on profiling them, so it should end up closer together than it is now.

TSGut:
Factor 2 speed-up bringing Q closer to R method. Let's see if there is anything left to do (which doesn't involve new custom struct to replace SymTridiagonal which is always an option).

XraisingLegendreImproved

TSGut:
Okay, I will leave it for now, I think the difference is cost has become acceptable considering the Q variant has to construct parts of the Householder matrices. The final comparison for a raising from Legendre looks like this.
XraisingLegendreImproved

Notably raising in some random Jacobi basis seems to put the two methods closer together, so maybe R has some additional advantages due to how Legendre is implemented somewhere. Here is a more general Jacobi example:
XraisingJacobiImproved

In any case they are both easily fast enough to go far beyond 100k entries so we probably don't need to worry too much. I just didn't like the 5x cost, 1.5x is something I can stomach given that the Q approach is more involved.

MikaelSlevinsky:
Let's say it's fast enough! There's probably a hard theoretical limit where the Q method is some specific constant slower than the R method.

Are you computing both bands of the modified Jacobi matrix independently? Looks like some code specializes on :dv and :ev.

TSGut:
Yes but `R` and `Q` both do this, which makes both of them at minimum a factor 2 slower than their theoretical limit since they have to redo the computations for each band. The reason for that is that we want this to be a cached `SymTridiagonal` which takes two bands as input. To be consistent with `SymTridiagonal` , Julia needs to know how to extend each band individually and since each band is its own object it can't tell Julia to simultaneously extend the other one.

The way around this would be to make a new struct as a subset of AbstractCachedArray which then gets extended as a whole entity. This would be a factor 2 speed-up for both methods but I had it set up that way before and after Sheehan's feedback in the previous PR we settled on it being a SymTridiagonal.

This would be easy to do if we want to tease out more efficiency but it would "fit" less well into existing packages which expect `SymTridiagonal` for a jacobimatrix.

TSGut:
I should add: There are certain advantages to the current approach. Resizing a vector is better than resizing a matrix, so if we resize often at small dimensions (arguably the usual use-case) this will be faster. But to reach high `N` in one swoop, this approach underperforms the theoretical limit by a factor 2.

MikaelSlevinsky:
Maybe the key then is for both bands to store the same cached QR factorization of sqrtw? This could be done I think if the cached qr(sqrtw) could be called in qr_jacobimatrix and it gets passed to both band constructors.

TSGut:
That would work, yes. It wouldn't affect the factor 2 I spoke of since that's all the householder stuff we have to redo but it should lead to a speedup nonetheless. Should be a quick change.

TSGut:
As a matter of fact, the Cholesky version already does this. So that's an oversight in the QR variant.

TSGut:
I am somewhat surprised that this doesn't appear to make a measureable performance impact but it still makes sense to do it this way regardless. I will check tomorrow whether I missed something.

TSGut:
I'll take another stab at reducing the cost in a bit.

TSGut:
Ok, so I of course wasn't computing dense householder matrices per se, I was computing the dense acting block of the householder which is an `O(b)` where `b=bandwidths(sqrtW)` operation and allocation, so it's cheap.

I have benchmarked it now and pre-allocating the `b x b` part of `H` (without ever densely computing `H`) is faster for computing all the required entries in the block `H*M*H` where `M` is a `O(b) x O(b)` block that moves down the band as we expand. The reason it's cheaper to pre-allocate the important part of `H` I think is that writing it all out the action of `H*M*H` is something like `(M-vv'M-Mvv'+vv'Mvv'` (I am omitting the tau constants for simplicity). And we really do need all those entries, not just one of them, to be adaptive. Doing it in this way unnecessarily recomputes the `vv'` repeatedly which in benchmarks costs more than just pre-computing it and re-using it (and this is equivalent to computing the acting block of `H`). So I don't think this path leads to any more speed-ups. To be clear, the `v`, `M` and `H` are all just of sizes `b` or `b x b` so there is very little to gain here either way.

Nevertheless, this tip _did_ lead me to find an operation that was `O(n*b^2)` for no reason and reduce it to `O(n*b)`, so `Q` is even closer to `R` now. The general Jacobi case is now strikingly similar:
XraisingJacobiImproved

I think it's fair to say they now perform de-facto the same. In any case these optimizations are mainly for vanity I think, or so that we can say in the paper that the two methods perform similarly well.

TSGut:
@dlfivefifty I am happy for this to be merged if all tests pass.

src/choleskyQR.jl
```diff
@@ -114,49 +114,94 @@ end
...
+function QRJacobiBand{:dv,:Q}(F, P::OrthogonalPolynomial{T}) where T
+    b = 3+bandwidths(F.R)[2]÷2
+    X = jacobimatrix(P)
+        # we fill 2 entries on the first run
+    dv = zeros(T,2)
+        # fill first entry (special case)
+    M = X[1:b,1:b]
+    getindex(F.factors,b,b) # pre-fill cached array
+    v = [one(T);F.factors[1:b,1][2:end]]
+    H = I - F.τ[1]*v*v'
```
dlfivefifty, line 154: This is definitely a dense Householder.... might be small dimensional but you are still forming it first and doing `H*M` instead of just doing the cheaper (mostly) in-place:
```julia
M .= M .- F.τ[1] .*  v .* (v'M)
M .= M .- F.τ[1] .*  (M*v) .* v'
```

dlfivefifty (changes requested):

src/choleskyQR.jl
```diff
@@ -114,49 +114,94 @@ end
...
+    M = H*M*H
+    dv[1] = M[1,1]
+        # fill second entry
+    v = [zero(T);one(T);F.factors[1:b,2][3:end]]
+    H = I - F.τ[2]*v*v'
```
dlfivefifty, line 159: Below too

src/choleskyQR.jl
```diff
@@ -171,23 +216,72 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
+    X = jacobimatrix(J.P)[1:m+b+2,1:m+b+2]
+    getindex(J.U.factors,m+b,m+b)
+    getindex(J.U.τ,m)
+    M = J.UX
+    τ = J.U.τ
+    F = J.U.factors
+    dv = J.data
+    @inbounds for n in jj
+        v = [zero(T);one(T);F[n-1:n+b,n][3:end]]
+        H = I-τ[n]*v*v'
```
dlfivefifty, line 232: Here too. This is the most expensive case.

src/choleskyQR.jl
```diff
@@ -171,23 +216,72 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
+        K = H*M*H
+        M = Matrix(X[n:n+b+1,n:n+b+1])
```
dlfivefifty, line 234: Pre-allocate `M` here. It's not a good idea to be allocating memory inside a loop

src/choleskyQR.jl
```diff
@@ -171,23 +216,72 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
+    m, jj = inds[end], inds[2:end]
+    X = jacobimatrix(J.P)[1:m+b+2,1:m+b+2]
+    getindex(J.U.factors,m+b,m+b)
+    getindex(J.U.τ,m)
+    M = J.UX
+    τ = J.U.τ
+    F = J.U.factors
+    dv = J.data
+    @inbounds for n in jj
+        v = [zero(T);one(T);F[n-1:n+b,n][3:end]]
```
dlfivefifty, line 231: Pre-allocate `v`

src/choleskyQR.jl
```diff
@@ -171,23 +216,72 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
+    end
+    J.UX = M
+    J.data[jj] = dv[jj]
+end
+function cache_filldata!(J::QRJacobiBand{:ev,:Q,T}, inds::UnitRange{Int}) where T
+    m, jj = 1+inds[end], inds[2:end]
+    b = bandwidths(J.U.factors)[2]÷2
+    # pre-fill cached arrays to avoid excessive cost from expansion in loop
+    X = jacobimatrix(J.P)[1:m+b+2,1:m+b+2]
+    getindex(J.U.factors,m+b,m+b)
```
dlfivefifty, line 246:
```suggestion
    resizedata!(J.U.factors,m+b,m+b)
```

src/choleskyQR.jl
```diff
@@ -171,23 +216,72 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
+    getindex(J.U.R,m,m)
```
dlfivefifty, line 247:
```suggestion
    resizedata!(J.U.R,m,m)
```

src/choleskyQR.jl
```diff
@@ -171,23 +216,72 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
+    getindex(J.U.τ,m)
```
dlfivefifty, line 248:
```suggestion
    resizedata!(J.U.τ,m)
````

File: src/choleskyQR.jl [hunk 9]
```diff
@@ -171,23 +216,72 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
+    M, τ, F, dv = J.UX, J.U.τ, J.U.factors, J.data
+    D = sign.(view(J.U.R,band(0)).*view(J.U.R,band(0))[2:end])
+    @inbounds for n in jj
```
dlfivefifty, line 251: prellocate `v` and `M` and `K`

src/choleskyQR.jl
```diff
@@ -171,23 +216,72 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
-    getindex(J.UX,inds[end]+1,inds[end]+1)
+    m = inds[end]+1
+    getindex(J.U,m,m)
+    getindex(J.UX,m,m)
 
     ek = [zero(T); one(T)]
+    dv, UX, U = J.data, J.UX, J.U
     @inbounds for k in inds
-        J.data[k] = dot(view(J.UX,k,k-1:k), J.U[k-1:k,k-1:k] \ ek)
+        dv[k] = dot(view(UX,k,k-1:k), U[k-1:k,k-1:k] \ ek)
```
dlfivefifty, line 271: Don't make a 2x2 matrix and then use `\`

src/choleskyQR.jl
```diff
@@ -171,23 +216,72 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
     end
+    J.data[inds] = dv[inds]
 end
-function cache_filldata!(J::QRJacobiBand{:ev, T}, inds::UnitRange{Int}) where T
+function cache_filldata!(J::QRJacobiBand{:ev,:R, T}, inds::UnitRange{Int}) where T
     # pre-fill U and UX to prevent expensive step-by-step filling in of cached U and UX in the loop
-    getindex(J.U,inds[end]+1,inds[end]+1)
-    getindex(J.UX,inds[end]+1,inds[end]+1)
+    m = inds[end]+1
+    getindex(J.U,m,m)
```
dlfivefifty, line 278: resizedata! (and below)

src/choleskyQR.jl
dlfivefifty: That can't really be the case: micro-allocations like this are extremely expensive. And a fully in-place version should be 2x as fast:
```julia
julia> function inplaceH!(τ, v, M, w)
              mul!(w, M', v)
              M .= M .- τ .* v .* w
              mul!(w, M, v)
              M .= M .- τ .* w .* v'
       end
inplaceH! (generic function with 2 methods)

julia> τ = 4.5; v = randn(2); M = randn(2,2); w = similar(v);

julia> @btime inplaceH!(τ, v, M, w);
  119.387 ns (0 allocations: 0 bytes)

julia> @btime denseH(τ, v, M);
  203.525 ns (4 allocations: 384 bytes)
```

TSGut: Hm alright I'm convinced. I will get the changes done and check it again, maybe my PC was just doing something in the background, the timings are pretty short and thus volatile. Should use BenchmarkTools.

TSGut:
Ok I think this is done @dlfivefifty . Happy for it to be merged or reviewed again

dlfivefifty (changes requested):

src/choleskyQR.jl
```diff
@@ -171,23 +221,75 @@ function resizedata!(K::QRJacobiBand, nm::Integer)
...
+        K .= K .- τ[n] .*  (K*v) .* v'
+        M = Matrix(X[n:n+b+1,n:n+b+1])
+        M[1:end-1,1:end-1] .= K[2:end,2:end]
+        dv[n] = M[1,1] # sign correction due to QR not guaranteeing positive diagonal for R not needed on diagonals since contributions cancel
+        K .= M
+    end
+    J.UX = M
+    J.data[jj] .= dv[jj]
+end
+function cache_filldata!(J::QRJacobiBand{:ev,:Q,T}, inds::UnitRange{Int}) where T
```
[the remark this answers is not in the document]
TSGut: I had it be in tandem originally in the previous PR but you didn't like my extra struct to accomplish it. I can definitely do that by introducing a helper struct that stores both bands. Shouldn't take long and like I discussed with mikael, this is an obvious 2x speed up.

src/choleskyQR.jl
```diff
@@ -90,73 +90,123 @@ function resizedata!(K::CholeskyJacobiBand, nm::Integer)
...
+    U = F.R
+    U = ApplyArray(*,Diagonal(sign.(view(U,band(0)))),U)  # QR decomposition does not force positive diagonals on R by default
     X = jacobimatrix(P)
     UX = ApplyArray(*,U,X)
     dv = zeros(T,2) # compute a length 2 vector on first go
-    dv[1] = dot(view(UX,1,1), U[1,1] \ [one(T)])
-    dv[2] = dot(view(UX,2,1:2), U[1:2,1:2] \ [zero(T); one(T)])
-    return QRJacobiBand{:dv,T}(dv, U, UX, 2)
+    dv[1] = UX[1,1]/U[1,1] # this is dot(view(UX,1,1), U[1,1] \ [one(T)])
+    dv[2] = dot(view(UX,2,1:2), [-U[1,2]/U[1,1],one(T)]./U[2,2]) # this is dot(view(UX,2,1:2), U[1:2,1:2] \ [zero(T); one(T)])
```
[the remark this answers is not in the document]
TSGut: can do

src/choleskyQR.jl
```diff
@@ -90,73 +90,123 @@ function resizedata!(K::CholeskyJacobiBand, nm::Integer)
...
+    b = 3+bandwidths(F.factors)[2]÷2
+    X = jacobimatrix(P)
+        # we fill 1 entry on the first run
+    dv = zeros(T,1)
+        # first step does not produce entries for the off-diagonal band (special case)
+    M = Matrix(X[1:b,1:b])
+    resizedata!(F.factors,b,b)
+    v = [one(T);F.factors[2:b,1]]
+    w = similar(v)
+    M .= M .- F.τ[1] .*  v .* (v'M)
```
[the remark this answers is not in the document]
TSGut: yeah that `w` was actually exactly from me experimenting with an in-place function like you say. I guess I will implement it that way then

src/choleskyQR.jl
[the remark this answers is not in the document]
TSGut: ok I think all your comments are done except this one - this might take a bit longer to make sure it doesn't introduce bugs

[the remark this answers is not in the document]
TSGut: Hm, maybe we should make sure on what's being considered before I restructure the whole thing, since it'd lead to redundant work.

If you recall, what I had originally was an adaptive $2\times \infty$ cached array that would construct both diagonal and off-diagonal in one swoop. That's nice bc it's fast but it's not that great bc SymTridiagonal takes vectors, not matrices and apparently resizing since vectors is a lot better than resizing matrices. Of course, the current method has the drawback that it recomputes for both bands (roughly a 2x cost increase) but it fits very nicely into SymTridiagonal.

So what is the best struct to go for here? Is there a way to get the best of both worlds or is it back to $2\times \infty$ cached array?

dlfivefifty:
Maybe `QRJacobiData` just stores two vectors that it resizes in tandem?

src/choleskyQR.jl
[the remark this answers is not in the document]
dlfivefifty: Maybe QRJacobiData just stores two vectors that it resizes in tandem?

TSGut:
Ok, I reworked it such that both bands are generated together while retaining the nice interaction with SymTridiagonal. This is a strange object so there may still be some stuff to tidy further but it all works at least.

This hasn't affected the gap between R and Q, which is still about as narrow as shown above but it has given us an approximately 2x speed-up (it's less than 2x because we were already resizing a joint QR / Cholesky so the workload was already partially shared).

src/choleskyQR.jl
```diff
@@ -51,143 +51,200 @@ function cholesky_jacobimatrix(W::AbstractMatrix, Q)
...
-    getindex(J.UX,inds[end]+1,inds[end]+1)
+    resizedata!(J.U,inds[end]+1,inds[end]+1)
+    resizedata!(J.UX,inds[end]+1,inds[end]+1)
 
-    ek = [zeros(T,2); one(T)]
+    dv, ev, UX, U = J.dv, J.ev, J.UX, J.U
     @inbounds for k in inds
-        J.data[k] = dot(view(J.UX,k,k-1:k+1), J.U[k-1:k+1,k-1:k+1] \ ek)
+        # this is dot(view(UX,k,k-1:k), U[k-1:k,k-1:k] \ ek)
+        dv[k] = -U[k-1,k]*UX[k,k-1]/(U[k-1,k-1]*U[k,k])+UX[k,k]./U[k,k]
```
dlfivefifty, line 108:
```suggestion
        dv[k] = -U[k-1,k]*UX[k,k-1]/(U[k-1,k-1]*U[k,k])+UX[k,k]/U[k,k]
````

File: src/choleskyQR.jl [hunk 16]
```diff
@@ -51,143 +51,200 @@ function cholesky_jacobimatrix(W::AbstractMatrix, Q)
...
+        ev[k] = UX[k,k-1]/U[k+1,k+1]*(-U[k-1,k+1]/U[k-1,k-1]+U[k-1,k]*U[k,k+1]/(U[k-1,k-1]*U[k,k]))+UX[k,k]/U[k+1,k+1]*(-U[k,k+1]/U[k,k])+UX[k,k+1]/U[k+1,k+1]  
     end
+    J.dv[inds] .= dv[inds]
```
dlfivefifty, line 111: Why is this needed? `dv == J.dv` so this is copying the same data to istelf. Same with the next line.

src/choleskyQR.jl
```diff
@@ -51,143 +51,200 @@ function cholesky_jacobimatrix(W::AbstractMatrix, Q)
...
+    X = jacobimatrix(J.P)[1:m+b+2,1:m+b+2]
+    resizedata!(J.U.factors,m+b,m+b)
+    resizedata!(J.U.τ,m)
+    K, τ, F, dv, ev = J.UX, J.U.τ, J.U.factors, J.dv, J.ev
+    D = sign.(view(J.U.R,band(0)).*view(J.U.R,band(0))[2:end])
+    w = Vector{T}(undef,b+3)
+    M = Matrix{T}(undef,b+3,b+3)
+    @inbounds for n in jj
+        dv[n] = K[1,1] # no sign correction needed on diagonal entry due to cancellation
+        doublehouseholderapply!(K,τ[n+1],[zero(T);one(T);F[n+2:n+b+2,n+1]],w)
```
dlfivefifty, line 226: This is still allocating. Probably change `doublehoseholderapply!` to know that the first two entries of `v` are zero and 1.

dlfivefifty (changes requested):

src/choleskyQR.jl
TSGut: so it's passing the vcat vector here that is allocating?

dlfivefifty: this allocates twice: first by calling `F[n+2:n+b+2,n+1]` and second by concatenating a 0 and a 1

TSGut: ok I guess the first of these can be eliminated by passing a view instead. Then the function can assume, except for the n=1 special case, that the entries are 0 and 1.

TSGut: passing the view seems to be better but I have a feeling `mul!(w, M, [zero(T);one(T);v])` would still allocate? What's a sensible way around that which isn't rewriting matrix multiplication?

dlfivefifty: Probably just use `reflectorApply!`?

https://github.com/JuliaLang/julia/blob/3051f47a40c37a2e898b9814856f7d4173b6d8c4/stdlib/LinearAlgebra/src/generic.jl#L1595

TSGut: I have experimented with modifications of that, can't seem to get it to reduce allocations using that. I will keep trying for now.

That said, having benchmarked a lot of different variations, the current variant pushes allocations under the current `R` method implementation so that's nice.

```julia
julia> @time Rtimer(100000,f,P,sqrtW)
  0.618029 seconds (11.78 M allocations: 526.182 MiB, 22.24% gc time)
0.617991553

julia> @time Qtimer(100000,f,P,sqrtW)
  0.654008 seconds (9.39 M allocations: 643.110 MiB, 19.61% gc time)
0.653952685
```

dlfivefifty: Did you look at `reflectorApply!`? You aren't using the QR data structure in the right way, you should be following existing coding patterns. (And of course the code is still allocating if you are calling `[zero(T); one(T); v]`.)

Note there is no need to concatenate a `1` because it just ignores the first entry. Eg the following code
```julia
    v = [one(T);view(F.factors,2:b,1)]
     mul!(w, M, v)  # M is symmetric
     M .= M .- F.τ[1] .* v .* w'
     mul!(w, M, v)
     M .= M .- F.τ[1] .* w .* v'
```
should just be written
```
    v = view(F.factors,1:b,1)
    reflectorApply!(v, F.τ[1], M)
    reflectorApply!(v, F.τ[1], M')
```
And the calls to `doublehouseholderapply!` similarly modified similarly, that is DO NOT concatenate a 0 and a 1 but instead call `reflectorApply!` with the right views of each piece.

dlfivefifty: Note this also means we don't even need the working variable `w`

TSGut: I tried something similar to that but ended up with more allocations, let's see if this template can get me there

TSGut: I think the part I may have been missing is that I can ignore the first entry, it may work like this

TSGut: I have to say I don't get `reflectorApply!(v, τ, M)`. I would have thought all it does is in-place left apply `I - τ*v*v'` to `M` and return this by overwriting `M` but evidently that is not what it does since using it in this way in the general case gives completely wrong entries.

dlfivefifty: It does `(I - τ*[1;v]*[1;v]')*M` in-place.

TSGut: Ah, I see. Then maybe I should first rewrite it into that style in my implementation, then it should be easy to just swap this function in.

TSGut:
@dlfivefifty Some mind bending indexing later, this now just uses `reflectorApply!`. I don't see any meaningful performance / allocation improvements (but to be fair also no deprecations, it's pretty much the same). But as said before the Q approach did dip under the R approach in allocations so perhaps there just isn't much more to be gained. Perhaps there is also some new sneaky allocation somewhere. I will take my eyes off it for a bit and have another look later today.

TSGut:
Ok, so @dlfivefifty:  apparently in 1.7 they changed `reflectorApply!()` to only accept matrices, i.e. it's function `reflectorApply!(x::AbstractVector, τ::Number, A::AbstractMatrix)`. This was then reverted at some point, in 1.9 at least it reads `function reflectorApply!(x::AbstractVector, τ::Number, A::AbstractVecOrMat)`. 

The 1.7 version in general seems to have had loss of functionality, so I am considering just copying the 1.9 version explicitly into this code. Alternatively we can make some special case for 1.7 but I guess I am not a big fan of that.

Your preference?

TSGut:
ok, I changed it to ask for 1.9, to only test on 1.9 and upped the version number since I guess this requirement is a significant change.

Other than that, I am once again happy with the state of the PR. ready for review or merge at your discretion assuming the tests pass

dlfivefifty (approved):

dlfivefifty:
I think its fine for now. There are still a few obvious allocations but its ok

PR #143, opened 2023-07-07
Lazy MemoryLayout for QR/Cholesky Jacobi matrices

TSGut:
Addresses the issue in https://github.com/JuliaApproximation/SemiclassicalOrthogonalPolynomials.jl/issues/77 and adds tests to check correct behavior


TSGut:
@dlfivefifty @ioannisPApapadopoulos ready for merge

PR #199, opened 2024-07-22
Implement gausslobatto / gaussradau

DanielVandH:
Implements `gaussradau` and `gausslobatto` for computing quadrature weights using Gauss-Radau or Gauss-Lobatto quadrature, respectively.

Is there a proper way I should be getting the monic polynomials? Or are my loops fine?

### DIFF
--- a/src/ClassicalOrthogonalPolynomials.jl
+++ b/src/ClassicalOrthogonalPolynomials.jl
@@ -318,6 +318,76 @@ end
 
 golubwelsch(V::SubQuasiArray) = golubwelsch(parent(V), maximum(parentindices(V)[2]))
 
+function gaussradau(P, n::Integer, endpt)
+    ## See Thm 3.2 in Gautschi (2004)'s book
+    # n is the number of interior points
+    J = symtridiagonalize(jacobimatrix(P, n + 1))
+    α = diagonaldata(J) 
+    β = supdiagonaldata(J) 
+    T = eltype(P) 
+    endpt = T(endpt)
+    p0 = zero(T) 
+    p1 = one(T)
+    for i in 1:n 
+        # Evaluate the monic polynomials πₙ₋₁(endpt), πₙ(endpt)
+        _p1 = p0 
+        p0 = p1 
+        p1 = (endpt - α[i]) * p0 
+        i > 1 && (p1 -= β[i - 1]^2 * _p1)
+    end 
+    a = endpt - β[end]^2 * p0 / p1 
+    α′ = vcat(@view(α[begin:end-1]), a)
+    J′ = SymTridiagonal(α′, β)
+    x, w = golubwelsch(J′)
+    w .*= sum(orthogonalityweight(P))
+    if endpt ≤ axes(P, 1)[begin]
+        x[1] = endpt # avoid rounding errors
+    else 
+        x[end] = endpt 
+    end
+    return x, w
+end
+
+function gausslobatto(P, n::Integer)
+    ## See Thm 3.6 of Gautschi (2004)'s book
+    # n is the number of interior points 
+    a, b = axes(P, 1)[begin], axes(P, 1)[end]
+    J = symtridiagonalize(jacobimatrix(P, n + 2))
+    α = diagonaldata(J) 
+    β = supdiagonaldata(J) 
+    T = eltype(P) 
+    a, b = T(a), T(b)
+    p0a = zero(T) 
+    p1a = one(T) 
+    p0b = zero(T) 
+    p1b = one(T) 
+    for i in 1:(n+1)
+        # Evaluate the monic polynomials πₙ₋₁(a), πₙ₋₁(b), πₙ(a), πₙ(b)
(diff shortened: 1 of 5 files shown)


dlfivefifty:
There is a routine `forwardrecurrence!` that builds OPs fast. The benefit over the for-loop is it special cases `Zeros`, `Vcat`, etc. to get high-efficiency. Whereas your for-loop will need to do slow checks.

But the cost of building the OPs is so fast compared to computing eigenvalues it wouldn't matter anyways.

Though perhaps its worth making a `Monic` type a la `Normalized` to just abstract away this step?

DanielVandH:
Would you prefer I rewrite it using `forwardrecurrence!`? I didn't know about that function. If not maybe coming back to `Monic` eventually and rewriting this would be a good approach too

dlfivefifty:
I think making a type for ` Monic ` and just using your for loop implementation for ` getindex ` with a comment about rewriting in the future would be good. This way if we need monic OPs for something in the future we remember that code is there!

DanielVandH:
Alright. I'll add `Monic` to this PR (or I can make a second PR if you prefer?) and reuse it, good idea.

DanielVandH:
Now added `Monic`. Just added the basics for now and I guess if/when another need for it comes up we can figure out the details a bit better

src/ClassicalOrthogonalPolynomials.jl
```diff
@@ -318,6 +319,51 @@ end
...
 
+function gaussradau(P::Monic{T}, n::Integer, endpt) where {T}
+    ## See Thm 3.2 in Gautschi (2004)'s book 
+    # n is number of interior points 
+    J = jacobimatrix(P.P, n + 1)
+    α, β = diagonaldata(J), supdiagonaldata(J)
+    endpt = T(endpt)
+    p0 = P[endpt, n] # πₙ₋₁
+    p1 = P[endpt, n+1] # πₙ. Could be faster by computing p1 from p0 and πₙ₋₂, but the cost is tiny relative to eigen()
+    a = (endpt - β[end]^2 * p0 / p1)::T
```
dlfivefifty, line 330: Why is the `::T` needed?

src/monic.jl
```diff
@@ -0,0 +1,42 @@
...
+axes(P::Monic) = axes(P.P)
+
+orthogonalityweight(P::Monic) = orthogonalityweight(P.P)
+
+_p0(::Monic{T}) where {T} = one(T)
+
+show(io::IO, P::Monic) = print(io, "Monic($(P.P.P))")
+show(io::IO, ::MIME"text/plain", P::Monic) = show(io, P)
+
+function getindex(P::Monic{T}, x::Number, n::Int)::T where {T}
```
dlfivefifty, line 23: Actually is it easier to just overload `recurrencecoefficients`? That will call `forwardrecurrence!` automatically.

src/ClassicalOrthogonalPolynomials.jl
```diff
@@ -318,6 +319,51 @@ end
...
 golubwelsch(V::SubQuasiArray) = golubwelsch(parent(V), maximum(parentindices(V)[2]))
 
+function gaussradau(P::Monic{T}, n::Integer, endpt) where {T}
+    ## See Thm 3.2 in Gautschi (2004)'s book 
+    # n is number of interior points 
+    J = jacobimatrix(P.P, n + 1)
+    α, β = diagonaldata(J), supdiagonaldata(J)
+    endpt = T(endpt)
+    p0 = P[endpt, n] # πₙ₋₁
+    p1 = P[endpt, n+1] # πₙ. Could be faster by computing p1 from p0 and πₙ₋₂, but the cost is tiny relative to eigen()
```
dlfivefifty, line 329:
```suggestion
p0, p1 = P[endpt,n:n+1]
```
If we overload `recurrencecoefficients` below this will automatically be fast.

dlfivefifty (changes requested):

src/ClassicalOrthogonalPolynomials.jl
DanielVandH: To help the inference a bit, since `β[end]^2` is inferred as `Any`. I can probably remove it since the whole function doesn't get inferred correctly anyway.

dlfivefifty: oh thats weird

DanielVandH:
Anything left to do here?

src/monic.jl
```diff
@@ -0,0 +1,38 @@
+struct Monic{T,OPs<:AbstractQuasiMatrix{T},NL} <: OrthogonalPolynomial{T}
+    P::Normalized{T,OPs,NL}
```
dlfivefifty, line 2: Does this actually benefit from requiring normalized OPs?

dlfivefifty (approved):

src/monic.jl
DanielVandH: Makes the code for getting the recurrence coefficients simpler I think

PR #214, opened 2024-11-26
Support DynamicPolynomials.jl

dlfivefifty:
This supports combining this package with symbolic computation. It's surprisingly fast! 3.5x faster than mathematica in fact:
```julia
julia> @time chebyshevt(10_000,x)
  0.078005 seconds (3.84 M allocations: 141.068 MiB, 18.68% gc time)
1 - 50000000x² + 416666650000000x⁴ - 5382805582903628800x⁶ - 2295839942961001984x⁸ - 5881664067181658112x¹⁰ + 9071434676160905216x¹² + 2726131615473205248x¹⁴ - 8311353073181097984x¹⁶ - 7478286491391623168x¹⁸ - 6167377969439309824x²⁰ - 749395590768492544x²² - 3810475745168326656x²⁴ + 5035205751279190016x²⁶ + 5745040218969341952x²⁸ - 5921272124512665600x³⁰ + 3935048435299778560x³² + 5526248795295186944x³⁴ - 4574133298292326400x³⁶ + 379621782652452864x³⁸ - 7095975366782615552x⁴⁰ - 9126263169889599488x⁴² - 5118059501529858048x⁴⁴ + 6924284427082137600x⁴⁶ + 7205196453839372288x⁴⁸ + 1729382256910270464x⁵⁰ + 2882303761517117440x⁵² - 4611686018427387904x⁵⁴ + 6917529027641081856x⁵⁶
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClassicalOrthogonalPolynomials"
 uuid = "b30e2e7b-c4ee-47da-9d5f-2c5c27239acd"
 authors = ["Sheehan Olver <solver@mac.com>"]
-version = "0.13.7"
+version = "0.13.8"
 
 [deps]
 ArrayLayouts = "4c555306-a7a7-4459-81d9-ec55ddd5c99a"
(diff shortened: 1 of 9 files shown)


blegat:
JuMP expressions are also `AbstractMutable`, I'm wondering if this works now:
```julia
using JuMP
model = Model()
@variable(model, x)
chebyshevt(10_000, x)
```

dlfivefifty:
The offending line is here:

https://github.com/JuliaApproximation/RecurrenceRelationshipArrays.jl/blob/bac197ddaa3bd5a1fbdb0f838d98e8658d21a734/src/recurrence.jl#L30

You can see it makes the mistake of assuming the output type is just the promotion of the input types.  Probably a better design might be:
```julia
    p1 = muladd(A[1],x,B[1])*p0
    T = typeof(p1)
     p0 = convert(T, μ)
```
Though some thought would be needed for the `N == 0` special case...

blegat:
You probably don't want the type of the result to depend on `N` so it's fine if you use the type corresponding for larger `N`

Issue #242, opened 2025-08-07
(If in scope) add extrema


eprovst:
It would be useful to provide implementations for `minimum`, `maximum` and `extrema`. I guess (going by Chebfun) this would effectively boil down to:
```julia
extrema(f) = extrema(map(x->f[x], [-1; findall(iszero, diff(f)); 1]))
```
or something similar.

dlfivefifty:
I'll add this. Note that you can just do:
```julia
extrema(f) = extrema(f[[-1; findall(iszero, diff(f)); 1]]))
```

PR #246, opened 2025-09-29
sampling, extrema (#242) and preliminary support orthogonalising quasimatrices

dlfivefifty:


dlfivefifty:
The test failure is unrelated

